Issue #7, opened 2016-07-16
Implement `get_row` methods for Matrix and MatrixSlices


AtheMathmo:
It would be nice if we could implement a `get_row` method to get a slice of a single row from a Matrix. Something like the following:

```rust
impl Matrix<T> {
    pub fn get_row(&self, idx: usize) -> &[T] {}

    pub fn get_row_mut(&mut self, idx: usize) -> &mut [T] {}
}
```

The contents of the method will need to be unsafe as we'll have to play with pointers a little within the slices. The above methods should do safety checks to ensure the indices are within the right range.

It may also be useful to implement unsafe `get_row_unchecked` and `get_row_unchecked_mut` which can be reused in the above methods.

The code in the [`iter_rows` methods](https://github.com/AtheMathmo/rulinalg/blob/054d7361d38401c500b46f68e526c9a468338655/src/matrix/iter.rs) can likely be adapted to produce the `get_row` methods etc.

tafia:
I think we should probably follow rust `slice` style (with a `Some`). 
```rust
pub fn get_row(&self, idx: usize) -> Some(&[T]) {}
pub fn get_row_mut(&self, idx: usize) -> Some(&mut [T]) {}
unsafe pub fn get_row(&self, idx: usize) -> &[T] {}
unsafe pub fn get_row_mut(&self, idx: usize) -> &mut [T] {}
```
What do you think?

AtheMathmo:
Yes I think you are right. It seems that `get` methods should return an `Option` if they may fail to meet a criteria.

_Minor note: the unsafe method names should be `get\_row\_unchecked_`

tafia:
Yes of course
I'll do it then, probably today

AtheMathmo:
That would be awesome! If you need any help feel free to reach out here or on [gitter](https://gitter.im/AtheMathmo/rusty-machine).

PR #8, opened 2016-07-26
implement get_row_* for Matrix, MatrixSlice and MatrixSliceMut

tafia:
- get_row
- get_row_mut
- get_row_unchecked
- get_row_mut_unchecked

Note:
For MatrixSlices, get_row and get_row_unchecked are implemented
directly on BaseSlice trait, as it already provides all necessary methods.

Similarly, moved get_unchecked slice implementations directly into the
trait, I don't really know but this may be a breaking change (I can revert this part if necessary).

### DIFF
--- a/src/matrix/mod.rs
+++ b/src/matrix/mod.rs
@@ -285,12 +285,11 @@ impl<T> Matrix<T> {
     /// assert!(a.get_row(5).is_none());
     /// ```
     pub fn get_row(&self, index: usize) -> Option<&[T]> {
-
-        if index >= self.rows {
-            return None;
+        if index < self.rows {
+            unsafe { Some(self.get_row_unchecked(index)) }
+        } else {
+            None
         }
-
-        unsafe { Some(self.get_row_unchecked(index)) }
     }
 
     /// Returns the row of a `Matrix` at the given index without doing unbounds checking
@@ -327,12 +326,11 @@ impl<T> Matrix<T> {
     /// assert!(a.get_row_mut(5).is_none());
     /// ```
     pub fn get_row_mut(&mut self, index: usize) -> Option<&mut [T]> {
-
-        if index >= self.rows {
-            return None;
+        if index < self.rows {
+            unsafe { Some(self.get_row_unchecked_mut(index)) }
+        } else {
+            None
         }
-
-        unsafe { Some(self.get_row_unchecked_mut(index)) }
     }
 
     /// Returns a mutable reference to the row of a `Matrix` at the given index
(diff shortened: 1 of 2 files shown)


tafia:
close #7

AtheMathmo:
Overall this is really good - just a few minor things.

Instead of

```rust
fn get_row(&self, index: usize) -> Option<&[T]> {

    if index >= self.rows() {
        return None;
    }

    unsafe { Some(self.get_row_unchecked(index)) }
}
```

I think it would be more rust like to use:

```rust
fn get_row(&self, index: usize) -> Option<&[T]> {
    if index < self.rows() {
        unsafe { Some(self.get_row_unchecked(index)) }
    } else {
        None
    }
}
```

I swapped the comparison out of personal preference - I wont hold it against you if you switch back :).

And also I'd appreciate if you can add some tests too. Just check that the unchecked version return the correct row in a basic test. And then check that the checked version also returns a valid `Some` option, or a `None` when applicable. Also testing that mutability works for `get_row_mut` would be appreciated - I've broken that a few times when I've moved things around so try and remember to add tests now... If you don't have time I'm happy to add these myself.

The actual code itself here is good (I believe) - I just want the tests to help prevent breakage if we move things around in future.

> Similarly, moved get_unchecked slice implementations directly into the
trait

You're right - I think this is a good change!

AtheMathmo:
Actually - I just noticed that you already have tests in the documentation to cover most things I mentioned. Sorry for the oversight.

If you could add them just to check that `Some`/`None` are returned where appropriate that would be greatly appreciated.

tafia:
Done.

>  I just noticed that you already have tests in the documentation to cover most things

Yes. The only missing part was for the `BaseSlice` implementations (there were no examples for the other functions).
Anyway added some example here too.

tafia:
I have another doubt: shoudn't we have:
```rust
fn get_row(&self, index: usize) -> Option<MatrixSlice>;
fn get_row_mut(&self, index: usize) -> Option<MatrixSliceMut>;
```

AtheMathmo:
This question has come up before and I opted not to do this.

The main reasoning is that &[T] guarantees contiguous data which allows us to optimize some operations. This isn't true of MatrixSlice. Right now I haven't run into a case where I need to treat a row as a matrix in itself (I think). Maybe I'll change this later though.

I'm happy to merge this now - but I'll give you some time to respond in case you disagree.

tafia:
I'm fine.
I think the best would be to have kind of a VectorSlice instead (that
guarantees continuous memory as well).
VectorSlice: Deref<[T]> or something like that.
But this would be a too big change, let's do this later if truly needed.

AtheMathmo:
Yes I agree - that may be a reasonable way to achieve this. I'll merge now and release this as 0.2.1 shortly.

Thanks again!

PR #11, opened 2016-07-26
Make vector implement Hash trait

### LINKED ISSUE #10: Make vector hashable
Vector does not implement the Hash trait, which makes it - for example - unusable as the key of a HashMap

brendan-rius:
Fix #10

### DIFF
--- a/src/matrix/mod.rs
+++ b/src/matrix/mod.rs
@@ -33,7 +33,7 @@ pub enum Axes {
 /// The `Matrix` struct.
 ///
 /// Can be instantiated with any type.
-#[derive(Debug, PartialEq)]
+#[derive(Debug, PartialEq, Hash)]
 pub struct Matrix<T> {
     rows: usize,
     cols: usize,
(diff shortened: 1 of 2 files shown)


AtheMathmo:
I think that it probably is worth adding. But could you please humor me and compare the compile times with and without `Hash` derived on `Matrix` and `Vec`? If it isn't significant I'll be happy to accept PR #11 .

If the compile time does increase dramatically for whatever reason then you should still be able to build a hashmap. You will just have to wrap access to the hashmap with methods building new `Vectors` (`Vector::new(key)`) and taking ownership of the data (`map.insert(vector.into_vec())`).

I feel that hashing is a rare use case but deriving isn't intrusive at all on the API or the code base (thanks Rust!). I'll be happy to include it if the cost is minimal.

brendan-rius:
On my computer, the original source code measured with

```bash
cargo clean && time cargo build
```

took 5.19s.

Adding `Hash` on Vector and Matrix made it go up to 5.28s.

This is clearly negligible

AtheMathmo:
I agree! Thanks for taking the time to check. And thank you also for the PR.

PR #13, opened 2016-07-29
Make Vector implement Eq trait

### LINKED ISSUE #12: Vectors do no support full equality
For any vector `v`, `v` is equal to itself.
This is then full equality and the vector needs to implement the trait `Eq`.
Right now vectors only implement `PartialEq`

Apart from logical reason to do this, some traits require `Eq` (for example the trait `Ord`).

brendan-rius:
Fix #12

### DIFF
--- a/src/matrix/mod.rs
+++ b/src/matrix/mod.rs
@@ -33,7 +33,7 @@ pub enum Axes {
 /// The `Matrix` struct.
 ///
 /// Can be instantiated with any type.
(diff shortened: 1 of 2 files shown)


AtheMathmo:
I think if we're adding this it is worth deriving the `Eq` for `Matrix` too. Would you mind adding that too? (Unless you have any objections)

AtheMathmo:
Thanks!

Issue #14, opened 2016-07-30
Implement Borrow and BorrowMut for Matrix


AtheMathmo:
This is a mirror to [rusty-machine/109](https://github.com/AtheMathmo/rusty-machine/issues/109).

AtheMathmo:
Some issues with implementing AsRef/Borrow:

```rust
impl<'a, T> AsRef<MatrixSlice<'a, T>> for Matrix<T> {
    fn as_ref(&self) -> &MatrixSlice<'a, T> {
        &self.as_slice()
    }
}

impl<'a, T> AsMut<MatrixSliceMut<'a, T>> for Matrix<T> {
    fn as_mut(&mut self) -> &mut MatrixSliceMut<'a, T> {
        &self.as_mut_slice()
    }
}
```

This code fails to compile with the following errors:

```
src/convert.rs:33:15: 33:23 error: cannot infer an appropriate lifetime for autoref due to conflicting requirements [E0495]
src/convert.rs:33         &self.as_slice()
                                ^~~~~~~~
src/convert.rs:32:5: 34:6 help: consider using an explicit lifetime parameter as shown: fn as_ref(&'a self) -> &MatrixSlice<'a, T>
src/convert.rs:32     fn as_ref(&self) -> &MatrixSlice<'a, T> {
                      ^
src/convert.rs:39:19: 39:31 error: cannot infer an appropriate lifetime for autoref due to conflicting requirements [E0495]
src/convert.rs:39         &mut self.as_mut_slice()
                                    ^~~~~~~~~~~~
src/convert.rs:38:5: 40:6 help: consider using an explicit lifetime parameter as shown: fn as_mut(&'a mut self) -> &mut MatrixSliceMut<'a, T>
src/convert.rs:38     fn as_mut(&mut self) -> &mut MatrixSliceMut<'a, T> {
                      ^
error: aborting due to 2 previous errors
error: Could not compile `rulinalg`.
```

But we cannot use an explicit lifetime as it goes against the trait:

```
error: method `as_ref` has an incompatible type for trait:
 expected bound lifetime parameter ,
    found concrete lifetime [E0053]
```

Do you see anything obviously wrong with this approach @tafia ? If not I'd propose we make our own internal traits: `AsSlice`and `AsMutSlice`. We can implement these for `Matrix`, `MatrixSlice` and `MatrixSliceMut` as appropriate - and use these over our generic functions. It should look something like this:

```rust
[...]

PR #18, opened 2016-08-02
DRY: Leverage more on BaseSlice trait

tafia:
We can leverage much more on trait to avoid repeating code and to simplify logic.

- Enhance `BaseSlice` (`iter`, ...), we should probably also have `iter_rows`, `select`etc ...
- Have `Matrix` implement `BaseSlice` (can answer issue #14 )
- Create a `BaseSliceMut` trait, which add mutability 
- Have `Matrix` and `MatrixSliceMut` implement `BaseSliceMut`

This is still a WIP, I think we should for instance be able to replace most arithmetic operations by something like (I cannot make it work yet):
```rust
impl<'a, T, S1, S2> Add<S1> for S2 
    where S1: BaseSlice<T>, S2: BaseSlice<T>,
{
// ...
}
```


tafia:
This is a breaking change as people need to import `BaseSlice` and `BaseSlice` trait most of the time.

AtheMathmo:
Thanks for putting so much effort into trying to solve this problem - I think this is another potentially powerful solution!

I want to go through and carefully review what you've done (I've only glanced over it so far) but I have some initial comments/questions about this approach.

I worry about putting a lot of functionality behind this trait and requiring the import. It adds some obscurity I think - perhaps we could go with the [`Float`](https://github.com/rust-num/num/blob/0faeb31b0963854168a5d6efaa7a7da099d3d25e/traits/src/float.rs) trait approach and simply call existing implementations within the trait and use it for generic-ness only?

I think having holdall traits like this can be quite risky - especially with an unstable API. If we rely on these for all generic behavior in future it could be hard to change. That said we're still <1.0.0 so we can chuck things around if *needed*.

I think using the trait more broadly is probably valuable outside of #14 . Helping reduce code bloat in arithmetic overloading in particular would be useful. Though bear in mind that the implementations for `Matrix` and `MatrixSlice` must be different as `Matrix` may use in place allocation if the matrix is consumed - while `MatrixSlice` may not.

Let me know what you think. I'll try to find some time today to read through the full PR (which may answer some of my above concerns).

And thanks again - this definitely feels like a step in the right direction.

tafia:
> I worry about putting a lot of functionality behind this trait and requiring the import. It adds some obscurity I think

I think this is how things work in a lot of rust crates (including std itself). The compiler will suggest the missing trait in its error message. Having a prelude-like module could help (even if I'm not a fan).

But it is really up to you, we can always add a macro that implements all of them, this is just an additional unnecessary step on compiler side. It feels less rust way to me.

tafia:
Just added `iter_rows(_mut)` into traits.
Had to moved `Rows` and `RowsMut` into `src/matrix/mod.rs`

AtheMathmo:
> I think this is how things work in a lot of rust crates (including std itself).

Yes you're right. I haven't been working with Rust for too long but maintaining these large traits with an unstable API seems painful. Maybe I'm wrong - in any case I think the benefits are probably worthwhile.

> Having a prelude-like module could help (even if I'm not a fan).

I'm not a fan either - but I have added on regardless!

I'll have to read through the PR in full before I can give any valuable feedback. It's looking like that may have to wait until tomorrow morning now - sorry.

I also have one more idea I want to try out with regards to getting AsRef/Borrow/Deref working. It may not work well in practice but I'll toy with that before making any decisions on this PR.

tafia:
No hurry!

tafia:
I don't manage to implement arithmetic operations using the traits because of the _orphan rule_ (from what I understand).

I can still modify the macros so it'll work for `MatrixSlice` and `Matrix` when possible.
By the way I've pushed changes so the `BaseSlice` uses associated types (`BaseSlice::Item` instead `BaseSlice<T>`): I think it is more flexible and easier to write in general.

I'll continue so DRY the code as much as possible.

AtheMathmo:
My other idea also came up short. It seems that we'll be waiting for [custom DSTs](https://github.com/rust-lang/rfcs/pull/1524) before we can get the ideal implementation,

I think this trait change will give us enough additional power for now anyway.

tafia:
I have moved most function like `transpose`, `hcat` etc... from `Matrix` to `BaseSlice`.

On one hand, I am sure there is no performance penalty, on the other, it brings more options to users (they don't need to say, convert a `MatrixSlice` into a `Matrix` to transpose it then).

I have also rewritten some of these functions to push an entire slice (via `get_row_unchecked`) instead of items one by one).

tafia:
Actually most fns in `src/matrix/mod.rs` and `src/matrix/decomposition.rs` can be moved to `BaseSlice` trait.

I think it makes sense, in particular if you want the crate to allow some parallel operations (it allows sharing parts of the matrix and working on them, via `MatrixSlice`s).

This will be quite a big change though (even if trivial), I don't know if you want this now or not. 

Anyway, I'm working on it, knowing that we can split this PR (first part until https://github.com/AtheMathmo/rulinalg/pull/18/commits/f214de80a4e3a58e563dc9bdc87c0fbfdec8c4b7, the rest on a second one) if needed.

AtheMathmo:
I think it makes sense to bring methods like `transpose` in which all matrices should implement. I had been holding back on a few of those (like `hcat`, `vcat`) because I feel as though there should be efficient ways to do this for the slices - but probably not.

As for the decompositions - it may be worth adding a new trait for these? I see no obvious benefit but it feels like this may help when things like specialization lands. This way we can implement the decomposition for `Complex` types differently (I think this works?).

As one final point I've been thinking it may be worth renaming this trait. `BaseSlice` isn't very descriptive of what the trait does anymore. I don't have any good suggestions yet, maybe: `Array` and `ArrayMut` as I think most of the behaviour captured here is specific to arrays? Then the arithmetic is defined on the Matrix structs themselves? I'm not super convinced by that idea so please feel free to recommend your own.

tafia:
> I feel as though there should be efficient ways to do this for the slices - but probably not.

Even if there are efficient ways, it should be a matter of overloading that particular function when we `impl BaseSlice for Matrix`.

> As for the decompositions - it may be worth adding a new trait for these? I see no obvious benefit but it feels like this may help when things like specialization lands. This way we can implement the decomposition for Complex types differently (I think this works?).

Looks like a good idea.

> I don't have any good suggestions yet, maybe: Array and ArrayMut as I think most of the behaviour captured here is specific to arrays?

I'm not sure exactly what is an `Array` (to me it is closer to a `Vector`, 1 dimension). What about `MatrixView(Mut)` ? The `View` part shows the intent of cheap borrow. 
Also I really feel like we could use a `VectorSlice/View`.

What is your  opinion about associated types in the trait. Does it feel right to you?

AtheMathmo:
> Even if there are efficient ways, it should be a matter of overloading that particular function

For sure - I just meant I had held off as I tried to figure those out. I think they should be put in now though.

You may be right about the `Array`. `MatrixView` might make more sense - but feels a little like a conflict with `MatrixSlice`. Are they different things? This trait to me seems to be capturing common behaviour amongst Matrix types, whether that's a view into the matrix or an owned matrix.

> Also I really feel like we could use a `VectorSlice`/`View`.

Yes I agree. I'll make a new issue for these.

As for associated types. It's hard to say as I still haven't properly read through and played with the PR. It doesn't seem necessary to me but maybe it is. In any case I think the name `Item` is a little misleading? It sounds too similar to the `Iterator` trait maybe?

tafia:
> It doesn't seem necessary to me but maybe it is

I actually agree. It looked like a good idea but it doesn't feel nice. Reverted it.

For the trait names, I'll change it later. Let me try adding a trait for decomposition / resolution first.

src/matrix/mod.rs
```diff
@@ -1021,175 +616,6 @@ impl<T: Float + FromPrimitive> Matrix<T> {
...
-                            "Matrix is not upper triangular.");
-                }
-
-            }
-        }
-
-        self.back_substitution(y)
-    }
-
-    fn back_substitution(&self, y: Vector<T>) -> Result<Vector<T>, Error> {
```
AtheMathmo: Some of these methods which have been pulled out should not be user facing. It's not that they are useless but rather they are not particularly well designed (yet) and so I had intentionally kept them for internal use only. The ones I saw are `back_substitution`, `forward_substitution` and `parity`. There may be some others.

src/matrix/slice.rs
```diff
@@ -86,44 +90,755 @@ pub trait BaseSlice<T> {
...
+            row_pos: 0,
+            slice_rows: self.rows(),
+            slice_cols: self.cols(),
+            row_stride: self.row_stride() as isize,
+            _marker: PhantomData::<&mut T>,
+        }
     }
+}
+
+impl<T> BaseSlice<T> for Matrix<T> {
```
AtheMathmo: I think this implementation would be best kept in mod.rs ? Or atleast it makes more sense to me if we rename the trait to something more `Matrix` focused.

src/matrix/slice.rs
```diff
@@ -487,7 +1061,7 @@ impl<'a, T: Copy> MatrixSliceMut<'a, T> {
     /// # Panics
     ///
     /// Panics if the dimensions of `self` and `target` are not the same.
-    pub fn set_to(self, target: MatrixSlice<T>) {
+    pub fn set_to(mut self, target: MatrixSlice<T>) {
```
AtheMathmo: Do we need this additional mut argument? I don't see what has changed? (Sorry if it is obvious)

AtheMathmo:
Had a slightly longer peek at the code. Just a few minor things - overall it looks good!

I'll wait for you to tackle decomposition before I really dive in. It may prove difficult as the current implementation works only for `Matrix`. If this is the case then we can spin it out into a new issue?

src/matrix/mod.rs
tafia: I can put them out of the trait, as standalone private functions (in `src/matrix/mod.rs` as this is used there as well). Would that be ok ?

src/matrix/slice.rs
tafia: If we don't put it we get:
```
src/matrix/slice.rs:1050:23: 1050:27 error: cannot borrow immutable argument `self` as mutable
src/matrix/slice.rs:1050         for (s, t) in self.iter_rows_mut().zip(target.iter_rows()) {
```

src/matrix/mod.rs
AtheMathmo: Yes I think that's the best approach. Though we'll need to make them generic so that we can use them for `Matrix`/`MatrixSlice`/`MatrixSliceMut`. I'm not sure how much of a challenge that will present?

tafia:
> If this is the case then we can spin it out into a new issue?

Let me try first. I'll let you know.

tafia:
>  I'm not sure how much of a challenge that will present?

Looks simple enough: https://github.com/AtheMathmo/rulinalg/pull/18/commits/2075fd7449560d9ff03466f9b92460025077084a

src/matrix/slice.rs
tafia: > I think this implementation would be best kept in mod.rs ? Or atleast it makes more sense to me if we rename the trait to something more Matrix focused.

Regarding that point, I'll do it at the end once we've settled a name

tafia:
I've added the `Decomposition` trait.
As much as possible I used `BaseSlice::into_matrix` which is a no-op for `Matrix`.

There are still some work to do I suppose but it should already be a good start.

tafia:
Pushed `inverse`, `det` and `solve` into `Decomposition` trait ... because it was easy and extend these functions to more items.

But we should probably rename (and relocate?) that trait then

tafia:
I think we can also merge the `Metric` trait into `BaseSlice` (or whatever the name)

AtheMathmo:
I'd like to keep the `Metric` trait out for now. I plan to extend it to some other norms later (general L-norms, etc.) and it would be nice to have these in one place (I think). For now at least...

I'm also not sure about having `inverse`/`det`/`solve` in `Decomposition`. Well - `inverse` may be ok but I feel that `solve` and `determinant` are not decompositions. Again this is something that may be worth readressing but for this PR I think it is best to keep them separate.

I agree with having `split_at` and `split_at_mut` in `BaseSlice` but I don't think that `Vector` should implement this? For example `hcat` should produce a `Matrix` but `vcat` should produce another Vector. The reason why `Vector` exists in this crate instead of just using a single column `Matrix` is that when we know the column has contiguous data we can do some things more efficiently. I haven't looked in detail but I worry some of this performance gain might be lost if we share the trait signatures. Or alternatively we'll have overly complex trait signatures.

Most of my comments here are things that I think might be worth exploring in the future but I'd like to see this PR land with a smaller subset of changes first and then address those things later.

If you disagree feel free to say so! You've been working with the code so may have seen things I'm missing.

AtheMathmo:
With regards to `Metric` - maybe we can implement for all types implementing BaseSlice? That makes more sense to me.

tafia:
> I'd like to keep the Metric trait out for now.

ok

> I'm also not sure about having inverse/det/solve in Decomposition. Well - inverse may be ok but I feel that solve and determinant are not decompositions. Again this is something that may be worth readressing but for this PR I think it is best to keep them separate.

I agree it is not `Decomposition` (this is why I thought it could be renamed as well). The point is that it is sad that it applies only on `Matrix` while it can easily be applied to slices as well. (I can definitely see think of situations where we prefer working with slices). An alternative is to have yet another trait or to have stand alone functions as well.

>  I don't think that Vector should implement this

I wasn't sure either. It doesn't bring anything on the table now. I'll revert.

> With regards to Metric - maybe we can implement for all types implementing BaseSlice? That makes more sense to me.

Yes. But we may get hurt by the orphan rule again.

AtheMathmo:
Hmm yes I see your point with regards to `inverse` etc. I'm not sure how best to handle it right now - hopefully I'll have a better idea after reading through. I think as you say moving those functions into the `Decomposition` trait might be the best option.

> Yes. But we may get hurt by the orphan rule again.

I think this wont be a problem as all traits are within the crate, but I could be wrong...
<br><br>

_Disclaimer on the following - my opinion may change after digesting more of the PR._

I think that some of these issues are why I'm a little cautious about moving things into traits like this. I agree it is the best way to go but leaves a lot to be desired compared to being able to use `Borrow` etc. Maybe the best approach is just to have one monolithic trait for all `Matrix` methods but this feels like an abuse of the type system to me - and one that I worry might come back to bite us later.

tafia:
>  I think as you say moving those functions into the Decomposition trait might be the best option.

:confused: I just pushed them into an `matrix::invertible::Invertible` trait ...

I'm leaving the `Metric` trait untouched for now.

src/matrix/slice.rs
```diff
@@ -38,90 +44,1094 @@ pub trait BaseSlice<T> {
...
 
-    fn row_stride(&self) -> usize {
-        self.row_stride
+impl<T> BaseSlice<T> for Matrix<T> {
+    fn rows(&self) -> usize { self.rows } 
+    fn cols(&self) -> usize { self.cols } 
+    fn row_stride(&self) -> usize { self.cols } 
+    fn as_ptr(&self) -> *const T { self.data.as_ptr() }
+
+    fn into_matrix(self) -> Matrix<T>
```
AtheMathmo: It's better to override `as_matrix` as this is the method which has the iteration and collection.

AtheMathmo:
> :confused: I just pushed them into an matrix::invertible::Invertible trait ...

Sorry for the confusion. This is a pretty huge PR so may take some time for things to settle in their ideal place :)

src/matrix/slice.rs
```diff
@@ -38,90 +44,1094 @@ pub trait BaseSlice<T> {
...
+    /// - The matrices have different column counts.
+    fn elemul<M>(&self, m: &M) -> Matrix<T>
+        where T: Copy + Mul<T, Output = T>,
+              M: BaseSlice<T>,
+    {
+        assert!(self.rows() == m.rows(), "Matrix row counts not equal.");
+        assert!(self.cols() == m.cols(), "Matrix column counts not equal.");
+
+        let mut data = Vec::with_capacity(self.rows() * self.cols());
+        data.extend(self.iter().zip(m.iter()).map(|(self_i, m_i)| *self_i * *m_i));
```
AtheMathmo: It may be worth checking how well this vectorizes for both `Matrix` and `MatrixSlice`. I think that using `iter_rows` and then the `utils::elemul` method may be better. And for `Matrix` we can take advantage of the fact that the whole data is contiguous.

When I first wrote these methods there was indeed a difference - but it seems the compiler has got a lot smarter with `zip` since then (as you pointed out). If we could validate this approach before making the change I'd appreciate it! (This would also probably warrant the removal of the utils `elemul`/`elediv` as they are only used here.

tafia: I've checked benchmarks. No regression on my system

src/matrix/slice.rs
tafia: as_matrix borrow self and return an owned Matrix. It is different than into_matrix which is a noop. We may override both though

src/matrix/slice.rs
AtheMathmo: Before I saw this I started testing it out on the playground and found that the zipped version doesn't vectorize :/ . It's possible that's an outdated rustc version or nightly interferig - but I doubt it.

[Code here](https://play.rust-lang.org/?gist=f10505e8dd54c18099b77670cd109a8e&version=nightly&backtrace=0)

If you go to release mode and compile to ASM you can see the difference. In the zipped mode the inner loop isn't vectorized (it uses [mulsd](http://docs.oracle.com/cd/E19253-01/817-5477/epmqg/index.html)). But with the elemul version the inner loop uses `mulpd` - which is multiplying packed values.

I'm by no means an expert on asm so I may be misunderstanding... When you ran the benchmarks did you do it on large data sizes? I don't think vectorization would make much difference unless the matrices are pretty hefty.

src/matrix/slice.rs
AtheMathmo: Ah I see! From memory I introduced this method as a hacky implementation of the `As` rust convention (`&T -> &U`) - where I couldn't have `&MatrixSlice -> &Matrix`. I think maybe it doesn't make that much sense though? I'm not sure when `as_matrix` would be valuable but `into_matrix` wouldn't be? Especially as `MatrixSlice` can be copied.

What do you think?

tafia: I think we should probably open an issue for that but keep it this way for the moment.
In theory both of them could be useful.

src/matrix/slice.rs
tafia: This is strange indeed.
I've ran existing benches, mat_mul_128_1000 in particular doesn't change.

```
$ rustc -V
rustc 1.12.0-nightly (34f35ed29 2016-07-17)

master
running 7 tests
test linalg::matrix::empty                  ... bench:           0 ns/iter (+/- 0)
test linalg::matrix::mat_create_100_100     ... bench:       2,695 ns/iter (+/- 71)
test linalg::matrix::mat_create_add_100_100 ... bench:       6,446 ns/iter (+/- 76)
test linalg::matrix::mat_mul_10_10          ... bench:         497 ns/iter (+/- 12)
test linalg::matrix::mat_mul_128_100        ... bench:     177,964 ns/iter (+/- 3,823)
test linalg::matrix::mat_mul_128_1000       ... bench:   1,695,926 ns/iter (+/- 41,726)
test linalg::matrix::mat_ref_add_100_100    ... bench:       5,520 ns/iter (+/- 173)

dry
running 8 tests
test linalg::matrix::empty                         ... bench:           0 ns/iter (+/- 0)
test linalg::matrix::mat_create_100_100            ... bench:       2,681 ns/iter (+/- 96)
test linalg::matrix::mat_create_add_100_100        ... bench:       6,602 ns/iter (+/- 170)
test linalg::matrix::mat_mul_10_10                 ... bench:         497 ns/iter (+/- 7)
test linalg::matrix::mat_mul_128_100               ... bench:     176,437 ns/iter (+/- 14,666)
test linalg::matrix::mat_mul_128_1000              ... bench:   1,684,683 ns/iter (+/- 374,040)
test linalg::matrix::mat_ref_add_100_100           ... bench:       5,373 ns/iter (+/- 214)
test linalg::matrix::mat_sum_rows_and_cols_128_100 ... bench:      15,790 ns/iter (+/- 508)
```

tafia:
I was also thinking to have a big wrapper trait that consumer could use instead of importing them one by one. This makes is cleaner on development side while remaining flexible on consumer side.

src/matrix/slice.rs
AtheMathmo: These benchmarks don't cover `elemul`/`elediv`. `mat_mul` is for normal matrix multiplication but here element wise multiplication is affected. Feel free to add benchmarks for these too if you like (I'll make an issue for it if not)

AtheMathmo:
I like the idea of a wrapper trait too. It definitely helps to alleviate some of my huge trait concerns.

tafia:
I've added benches and modified the code (there were a significant regression).

Here is a bench with the *maximum* penalty (row mod 8 = 7);

```
running 3 tests
test linalg::matrix::mat_elediv_63_1000            ... bench:      47,981 ns/iter (+/- 1,540)
test linalg::matrix::mat_elemul_63_1000            ... bench:      29,326 ns/iter (+/- 8,768)
test linalg::matrix::mat_mat_elemul_63_1000        ... bench:      26,098 ns/iter (+/- 721)
```
The mat_mat is the version `Matrix` * `Matrix`. As you can see there is a penalty in being too generic. I couldn't easily overload that function when implementing Matrix.
What do you propose ? Should we remove this or rename it differently ?

AtheMathmo:
I'm pretty sure the regression is caused by the fact that in the new `elemul` function we are vectorizing over each row - but we can vectorize over the entire data for `Matrix` (as it is all guaranteed to be contiguous).

If you add an overriding implementation for `Matrix` in the `BaseSlice` implementation you should see the performance improve again.

tafia:
Yes this is certainly where the regression comes from.

> If you add an overriding implementation for Matrix in the BaseSlice implementation you should see the performance improve again.

I tried but it didn't work. It can work being slightly less generic:
```rust
// overloadable if we remove <M>
fn elemul(&self, m: &Matrix) -> Matrix<T>
         where T: Copy + Mul<T, Output = T>,
               M: BaseSlice<T>,

// Error: I couldn't find a way to overload this version
//  error: wrong number of type arguments: expected 1, found 0 [E0243]
// src/matrix/slice.rs:1116     fn elemul(&self, m: &Matrix) -> Matrix<T>
fn elemul<M>(&self, m: &M) -> Matrix<T>
         where T: Copy + Mul<T, Output = T>,
               M: BaseSlice<T>,
```

AtheMathmo:
I'm away from home so can't test myself. But maybe we can try something like:

```fn elemul(&self, m: Self) -> Matrix<T>```

It isn't as generic as we'd like but is better than we had before. And users can get all desired behaviour by calling `as_slice` where needed.

I'm not sure that we can achieve exactly what we want without stooping to macros otherwise.

**Comment by tafia** [comment]
Let me do this

**Comment by tafia** [comment]
done.
Any remaining point I didn't cover?

**Comment by AtheMathmo** [comment]
It looks good to me from here but I'll have to give it a thorough read over on Tuesday. I want to make sure we don't miss anything as there's a lot here.

Thanks again for all the work you've done on this. I'm looking forward to making use of this!

**Comment by tafia** [comment]
Sure
Take your time

**Comment by tafia** [comment]
Pushed `as_slice` and `as_mut_slice`, just to be certain we can use `elemul` like functions easily (even multiplying `Matrix` and `MatrixSliceMut` for instance)

File: src/matrix/decomposition.rs [hunk 6]
```diff
@@ -516,17 +386,21 @@ impl<T: Any + Float + Signed> Matrix<T> {
     /// # Failures
     ///
     /// - The matrix cannot be reduced to upper hessenberg form.
-    pub fn upper_hessenberg(mut self) -> Result<Matrix<T>, Error> {
-        let n = self.rows;
-        assert!(n == self.cols,
+    fn upper_hessenberg(&self) -> Result<Matrix<T>, Error>
```
AtheMathmo: In the previous code we take `mut self` here. I think we should follow the same pattern you used above and take `self` and then call `into_matrix()`. 

This will be a noop for `Matrix` and the same cost as it currently is for slices.

AtheMathmo:
I've checked over the decomposition code. Other than my comment the rest looks good. In future I think it will be worth implementing these differently for each type of `Matrix` but we're not quite there yet.

I need to look over the other changes some more but from what I've seen they look good. I'll hopefully look in more detail tomorrow. Some comments for now though.

- We should describe the use of these traits in the crate and module documentation. Just explain that these traits contain the bulk of the behaviour and exist to allow code to be generic over each `Matrix` type. We don't need too much detail there.
- I'm not sure that we need `as_matrix` and `into_matrix`. As I mentioned in a previous comment I think that `as_matrix` was a mistake and doesn't fit the rust convention (`&T -> &U`). If you agree I think we should remove this and stick with `into_matrix`.

src/matrix/decomposition.rs
tafia: We can do this. The only issue is that there is ONE unnecessary `clone` in `francis_shift_eigenvalues`: [before](https://github.com/AtheMathmo/rulinalg/pull/18/files#diff-de7dc9f01b39718872f6eec6eb846335L696), [after](https://github.com/AtheMathmo/rulinalg/pull/18/files#diff-de7dc9f01b39718872f6eec6eb846335R911).
Is it ok to get this perf penalty (just as before) and getting rid of `as_matrix` ?

tafia:
> We should describe the use of these traits in the crate and module documentation

Yes

> I'm not sure that we need as_matrix and into_matrix

I understand the issue with the `as_matrix`. As I said, my only concern is that we need an unnecessary `clone` in one function. I think several parts of decomposition module could be written more efficiently using `BaseSlice` and `BaseSliceMut`.

> Before merging I think we should settle on the names of things

I totally agree. The changes are too big and will necessarily impact rulinalg users. Better to do it now.

tafia:
I also have another suggestion: 
- add `sub_slice(_mut)` functions which acts as the `reslice` but are more generic. It'll help in various part of the code where we use `from_matrix` or `from_raw_parts` ... 
- eventually deprecate the `reslice` functions as they will become redundant and are less explicit IMHO (no `mut` part in particular)

src/matrix/decomposition.rs
AtheMathmo: Ok I think that I understand - though this is a negative I think. The `as_matrix` method is essentially a clone (a less efficient one for `Matrix` right now - though we can override that). Having this clone in the `upper_hessenberg` method is unnecessary and makes that method less efficient (while the `eigenvalue` method is as inefficient as it was before - simply deferring the clone to later).

Also the decomp methods have additional clones right now. As we are calling `self_m.as_matrix()` and then cloning as we pass to `upper_hess_decomp()` and then within `upper_hess_decomp` we clone at the end when we call `upper_hessenberg` again.

I'm not sure if there is a good work around for this without rewriting large chunks of the decomposition code - which I really don't want to do in this PR as it is already a bit too large for me to handle :(. If you don't see an obvious fix maybe we should revert the decomposition changes and move these into a new PR? That way we can handle just the `BaseSlice` generic changes right now.

AtheMathmo:
I like the `sub_slice` suggestion. It seems like a good replacement for `reslice`.

I see why we need the `as_matrix` method but see my comment above. Perhaps I am misunderstanding some of what you've worked through during this PR. If we need to keep it then it's no big deal - having an extra function is hardly a deal breaker :).

My biggest concern right now is making sure we haven't sacrificed performance anywhere here (at least not where it can be helped). I know it may seem premature to worry about it but there's no way to be certain that anything we unknowingly introduce will get cleaned up any time soon.

> Do we keep the `Invertible` trait?

It does feel pretty clunky to me - if we get a better name for `Decomposition` than we can put it there. But I think otherwise it is a reasonable move for the sake of clarity (and with a wrapper trait will feel a little easier). Though see my comment above - I'm inclined to split out the `Decomposition` work into a new PR and keep them `Matrix` only for this PR. Would make it easier to review I think.

I should have some time to read through the other added modules late this evening.

tafia:
Reverted `Decomposition` ... and `Invertible: Decomposition`. I've saved the work on a `dry-backup` branch in my fork.

AtheMathmo:
Thanks, and it's good that you saved it. I think it's really valuable but probably needs more work and I'd like to get this merged soon.

I'll try to get through the rest of it in the next day and hopefully then we can get this in!

tafia:
I've also pushed `sub_slice` and kept `reslice` for the moment. I think we can remove it later.

tafia:
I will just change the trait names and it should be good.

AtheMathmo:
I saw, thanks! I agree we should leave reslice for now. I'll make an issue to remind me to remove it later though.

tafia:
Names changed.

> I saw, thanks! I agree we should leave reslice for now. I'll make an issue to remind me to remove it later though.

I copied the logic of reslice ... but I think there were a bug originally:
```rust
unsafe {
    self.ptr = self.ptr.offset((start[0] * self.cols + start[1]) as isize);
}
```
Shouldn't we offset the reslice with `row_stride` instead of `cols`:
```rust
unsafe {
    self.ptr = self.ptr.offset((start[0] * self.row_stride + start[1]) as isize);
}
```

AtheMathmo:
Ah - nice catch! My test only checked the first row and I haven't had a use case for `reslice` personally.

If you want to can you add a test to cover this case? If not (which is totally understandable) I'll write up an issue to add a test covering that case. Let me know.

tafia:
Done. The test was indeed not working before the fix.

AtheMathmo:
Looking through this and there are a few things I see. Most are minor and that I think we should address after the PR merges. A couple others I think we should address before merging (again - they are mostly minor).

To note after merging:

- `sum_rows` should probably use `iter_rows` and `utils::in_place_vec_bin_op`. I don't think this gives performance gains but will look nicer?
- `vcat` can be overridden for `Matrix` to be more efficient. We just concatenate the two data vectors instead of doing it per row. Maybe this should go in now but I'm happy to wait.

To do before merging:

- A lot of the `BaseMatrix` function documentation still refers to _slices_. I think we should change this to _matrices_? Happy to be overruled.
- Some documentation of `BaseMatrix` should be clearly visible in `lib`and `matrix/mod` as it is now responsible for a lot of functionality. Ideally we should also have something in the README but we can postpone that.
- Should we restructure this stuff? Having `BaseMatrix` in `slice` seems counter-intuitive. I'm happy to leave this for later but we should at least `pub use BaseMatrix` in `matrix/mod`.

What do you think of those? You've already spent a lot of time on this so if you like I can handle the last few things.

AtheMathmo:
@tafia could you let me know whether you want me to handle the last few tweaks? There's no rush if you want to.

Thanks!

tafia:
Yes I'll do it I have some work at the moment and I am not very good at
writing docs but I didn't forget. If you're in a hurry don't hesitate to do
it.

tafia:
I've done most of the changes in question. 

I have **not** 
- updated the README as I'm not clear on what should be there
- renamed the `slice` module, I've publicly imported the traits in `matrix/mod.rs` which should be a good start. (in particular, I'm wondering what will happen if we manage to update the decomposition module, will the names align?)
- added much documentation in `matrix/mod.rs`, I've merely added a reference in the module documentation saying that most logic is actually implemented on the traits.

Please review (and sorry for the time it took!)

AtheMathmo:
Thanks!

I think this will be enough to merge. I will handle updating the README (I just wanted to add an example illustrating this - or perhaps something even simpler).

I think `pub use` is enough for the traits - it is actually nicer as it keeps `matrix/mod` cleaner.

I'm away until Thursday so wont be able to review until then. A quick look through now and it looks ready to merge but as it is a large PR I want to give it one more thorough look through. Should have it in by Friday!

Thanks again

AtheMathmo:
Just finished looking this over and I'm pleased to say it's finally getting merged!

Thanks for all of your work on this! I'm excited to make use of these changes in rusty-machine.

tafia:
Thanks you!!

Issue #27, opened 2016-08-29
Improve documentation for sum_rows and sum_cols


AtheMathmo:
See #25 .

We should make it clear in the documentation how the sum is computed for each case.

PR #28, opened 2016-09-02
Improve documentation for sum_rows and sum_cols

scholtzan:
I updated the documentation of `sum_rows` and `sum_cols`. This addresses #27


AtheMathmo:
This is great - thanks!

PR #30, opened 2016-09-04
Generalise signature of select_rows and select_cols

theotherphil:
As discussed, I have a genuine use case for this in my cross_validate implementation.

If you think the extra complexity (in the signature at least, if not in the implementation) isn't worth it I'm happy to just replicate this internally to the cross validation module.

Sorry for the formatting changes mixed in - looks like my editor (or git?) automatically removes whitespace at the end of lines.

### DIFF
--- a/src/matrix/slice.rs
+++ b/src/matrix/slice.rs
@@ -280,7 +280,7 @@ pub trait BaseMatrix<T>: Sized {
 
         Matrix {
             cols: self.cols(),
-            rows: row_iter.clone().len(),
+            rows: row_iter.len(),
             data: mat_vec,
         }
     }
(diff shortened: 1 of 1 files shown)


AtheMathmo:
This looks good and I don't think the added complexity is bad at all. I'm happy to merge as is but I wonder if it's possible to remove the clones on the iterator? It shouldn't be an expensive operation but it feels unnecessary.

One alternative is to require I to implement `Copy`, then we can just use the `for x in _` syntax in each loop. I think the current approach is probably better though as a clone of the iterator is likely to be about as expensive as a `Copy` of the data (and is more generic).

I find it surprising there isn't a `BorrowedIterator` trait or something that exposes `iter()` - especially as so many `IntoIterator` seem to simply call `iter()`.

theotherphil:
The clones can't be gotten rid of, but I don't think they're a problem. Each clone just copies a fat pointer.

AtheMathmo:
Yes I agree. It is possible it would be more expensive but it would be an odd iterator and probably user created.

PR #31, opened 2016-09-05
Adding swap_rows and swap_cols for BaseMatrixMut

AtheMathmo:
Some rustfmt disagreement here but the core addition is the `swap_rows` function in the `BaseMatrixMut` trait. This function swaps the rows of a matrix or matrix slice in place without any allocation (using `mem::swap`).

Benchmarks on my system. These are 100x100 matrices.

```
test linalg::matrix::mat_swap_cols_0_99            ... bench:          60 ns/iter (+/- 1)
test linalg::matrix::mat_swap_rows_0_99            ... bench:          15 ns/iter (+/- 0)
```

### DIFF
--- a/benches/linalg/matrix.rs
+++ b/benches/linalg/matrix.rs
@@ -1,5 +1,5 @@
 use rulinalg::matrix::Matrix;
-use rulinalg::matrix::slice::BaseMatrix;
+use rulinalg::matrix::slice::{BaseMatrix, BaseMatrixMut};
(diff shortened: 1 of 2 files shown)


AtheMathmo:
Before merging this I'm going to add some benchmarks. I'm also going to explore a `swap_cols` function and see how it's performance compares.

It might be worth making this a single function that takes an `Axes` argument - but thinking of names for things is hard.

PR #32, opened 2016-09-06
Handle error cases for `mean` and `variance`.

### LINKED ISSUE #20: Have `mean` and `variance` methods return results
These methods can fail and it would be good for this to be recognized in the API.

The `mean` method may _fail_ if the input Matrix is empty - though it may make sense to simply return an empty mean vector in response.

The `variance` method may fail if the input Matrix has only one row/column of data in the working axis. In this case we will end up dividing by zero (as this method computes the unbiased [sample variance](http://mathworld.wolfram.com/SampleVariance.html)). For now we should fail in this case and perhaps later add in a new method for the biased sample variance which will simply return 0 in this case.

eugene-bulkin:
This resolves #20 with the following changes:

- Calling `mean` on an empty matrix returns an empty matrix vector.
- Calling `variance` on a matrix with one or fewer rows/columns in the working axis returns an `Error` with kind `InvalidArg`.

### DIFF
--- a/src/matrix/mod.rs
+++ b/src/matrix/mod.rs
@@ -334,11 +334,11 @@ impl<T: Float + FromPrimitive> Matrix<T> {
     /// // Only one column
     /// let b = Matrix::<f32>::new(2,1,vec![1.0,2.0]);
     ///
-    /// let e = b.variance(Axes::Row);
+    /// let e = b.variance(Axes::Row).unwrap();
     /// assert_eq!(*e.data(), vec![0.5]);
     ///
-    /// let f = b.variance(Axes::Col).unwrap();
-    /// assert!(*f.is_err());
+    /// let f = b.variance(Axes::Col);
+    /// assert!(f.is_err());
     ///
     /// ```
     /// // Empty matrix
@@ -347,7 +347,7 @@ impl<T: Float + FromPrimitive> Matrix<T> {
     /// let b = a.variance(Axes::Row);
     /// assert!(b.is_err());
     ///
-    /// let c = b.variance(Axes::Col);
+    /// let c = a.variance(Axes::Col);
     /// assert!(c.is_err());
     /// ```
     ///
--- a/src/matrix/mod.rs
+++ b/src/matrix/mod.rs
@@ -371,14 +371,10 @@ impl<T: Float + FromPrimitive> Matrix<T> {
             }
         }
 
-        if n == 0 {
+        if n < 2 {
             return Err(Error::new(ErrorKind::InvalidArg,
-                                  "There is no data in the working axis."));
-        }
-
-        if n == 1 {
-            return Err(Error::new(ErrorKind::InvalidArg,
-                                  "There is only one row or column in the working axis."));
+                                  "There must be at least two rows or columns in the working \
+                                   axis."));
         }
 
         let mut variance = Vector::zeros(m);
--- a/src/matrix/mod.rs
+++ b/src/matrix/mod.rs
@@ -322,6 +322,7 @@ impl<T: Float + FromPrimitive> Matrix<T> {
     /// ```
     /// use rulinalg::matrix::{Matrix, Axes};
     ///
+    /// // Only one row
     /// let a = Matrix::<f32>::new(1,2,vec![1.0,2.0]);
     ///
     /// let c = a.variance(Axes::Row);
@@ -330,13 +331,24 @@ impl<T: Float + FromPrimitive> Matrix<T> {
     /// let d = a.variance(Axes::Col).unwrap();
     /// assert_eq!(*d.data(), vec![0.5]);
     ///
-    /// let b = Matrix::<f32>::new(0,0,vec![]);
+    /// // Only one column
+    /// let b = Matrix::<f32>::new(2,1,vec![1.0,2.0]);
     ///
     /// let e = b.variance(Axes::Row);
-    /// assert!(e.is_err());
+    /// assert_eq!(*e.data(), vec![0.5]);
     ///
-    /// let f = b.variance(Axes::Col);
-    /// assert!(f.is_err());
+    /// let f = b.variance(Axes::Col).unwrap();
+    /// assert!(*f.is_err());
+    ///
+    /// ```
+    /// // Empty matrix
+    /// let a = Matrix::<f32>::new(0,0,vec![]);
+    ///
+    /// let b = a.variance(Axes::Row);
+    /// assert!(b.is_err());
+    ///
+    /// let c = b.variance(Axes::Col);
+    /// assert!(c.is_err());
     /// ```
     ///
     /// # Failures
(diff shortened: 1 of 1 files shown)


src/matrix/mod.rs
```diff
@@ -318,6 +359,16 @@ impl<T: Float + FromPrimitive> Matrix<T> {
             }
         }
 
+        if n == 0 {
```
AtheMathmo: I think we should capture this in one `if` block for brevity.

```rust
if n < 2 {
    ... "There must be at least two rows or columns in the working axis."
}
```

What do you think?

src/matrix/mod.rs
```diff
@@ -267,7 +267,24 @@ impl<T: Float + FromPrimitive> Matrix<T> {
...
+    /// let a = Matrix::<f64>::new(0,0, vec![]);
+    ///
+    /// let c = a.mean(Axes::Row);
+    /// assert_eq!(*c.data(), vec![]);
+    ///
+    /// let d = a.mean(Axes::Col);
+    /// assert_eq!(*d.data(), vec![]);
+    /// ```
     pub fn mean(&self, axis: Axes) -> Vector<T> {
+        if self.rows == 0 {
```
AtheMathmo: I think we should check here for self.cols too?

`if self.rows == 0 || self.cols == 0 { ... }`

AtheMathmo:
Thanks for getting this in! :D

I added a few inline comments. It would also be great if you could add some tests to cover the failure cases. If there are any non-doc tests we should make sure these are unwrapping too. I think the following cases should cover it:

These three can be covered in one test:

- Variance with empty data should return an error.
- Variance over row with 1 row should err.
- Same with column

- Matrix should return an empty Vector if matrix is empty.

src/matrix/mod.rs
eugene-bulkin: Yeah, I did them separately because initially I forgot about the `n = 0` case. Combining them seems reasonable to me.

src/matrix/mod.rs
eugene-bulkin: I prefer the latter, since `self.data.len() == 0` iff `self.rows == 0 || self.cols == 0`, and each individual operation should be only one instruction. (if that's not the case, we can do the first one)

eugene-bulkin:
I did not notice any non-doc tests for `mean` or `variance` so I only created doc tests.

I believe all but the "variance of one column" situation were in the doc tests, so I'll add that one along with the other fixes.

eugene-bulkin:
I think this should cover all the above changes. Let me know if I missed a spot.

AtheMathmo:
All of the code looks good but I'm afraid I'm going to have to be difficult, sorry!

I'd prefer if the failure test cases were in unit tests in that module rather than docs. When I first started I stuck a lot of tests in the docs but have tried to move away from that a little more to keep the examples easy to read.

I think it's enough to have one test each for matrix and vector. The matrix one can just test the empty matrix case returns an empty Vector. Maybe it's worth documenting this behavior in the matrix docs too?

Sorry again for asking you to do more than you already have.

eugene-bulkin:
> I'd prefer if the failure test cases were in unit tests in that module rather than docs. When I first started I stuck a lot of tests in the docs but have tried to move away from that a little more to keep the examples easy to read.

Not a problem, I was going to put it in `tests/mat/mod.rs` originally but I just wasn't 100% sure if you actually wanted them outside of docs.

> I think it's enough to have one test each for matrix and vector.

Not 100% sure what you mean by "one test each for matrix and vector", since aren't these just methods on `Matrix<T>`?

> The matrix one can just test the empty matrix case returns an empty Vector. Maybe it's worth documenting this behavior in the matrix docs too?

I was thinking the "empty mean" test would just test that case, yes. The variance test will be checking for errors in `n == 0` and both `n == 1` cases. As for documenting empty mean in the docs, yeah, I can toss a note in there noting that it does not fail, but rather returns an empty matrix.

AtheMathmo:
Sorry the "matrix and vector" was just my tiredness. I meant mean and variance.

eugene-bulkin:
How is this?

src/matrix/mod.rs
```diff
@@ -295,13 +307,33 @@ impl<T: Float + FromPrimitive> Matrix<T> {
...
+    /// let c = a.variance(Axes::Row).unwrap();
     /// assert_eq!(*c.data(), vec![2.0, 2.0]);
     ///
-    /// let d = a.variance(Axes::Col);
+    /// let d = a.variance(Axes::Col).unwrap();
     /// assert_eq!(*d.data(), vec![0.5, 0.5]);
     /// ```
-    pub fn variance(&self, axis: Axes) -> Vector<T> {
+    ///
+    /// ```
```
AtheMathmo: I meant to ask you to remove this part too. I don't mind keeping it here but if we do we should explain before presenting it that it is showcasing misuse of the function. A small sentence before the examples saying:

"Computing the variance will fail if the matrix has invalid dimensions:"

src/matrix/mod.rs
```diff
@@ -266,8 +268,18 @@ impl<T: Float + FromPrimitive> Matrix<T> {
     ///
     /// let d = a.mean(Axes::Col);
     /// assert_eq!(*d.data(), vec![1.5, 3.5]);
+    ///
+    /// let b = Matrix::<f64>::new(0,0, vec![]);
```
AtheMathmo: I meant to ask you to remove this too - but I think it isn't too intrusive and helps to explain the sentence above about calling `mean()` on empty.

src/matrix/mod.rs
eugene-bulkin: I believe the Failures section should cover that, no? That's what is used to denote reasons for a result to not be `Ok`, as far as I recall.

src/matrix/mod.rs
eugene-bulkin: I can get rid of it. I left it specifically for that reason, as an immediate example of the note in the docs.

src/matrix/mod.rs
AtheMathmo: I'm just being picky about the format :)

I just dislike the large example showing the errors as I think the **Failures** section details them well enough. It's not a big deal and if you think it is useful for the user I'll be happy to keep it in :).

eugene-bulkin: Nah, I'd be inclined to agree on that point. The Failures section should be sufficient to describe the issue.

src/matrix/mod.rs
eugene-bulkin: I think I'll get rid of it; not like it would be hard to put it back if someone complained that they needed an example.

AtheMathmo:
Thanks for putting up with my nitpicks. Once you've cleaned up the examples I'll be more than happy to get this merged!

eugene-bulkin:
Just pushed. Let me know if there's any other minor changes!

AtheMathmo:
This is perfect, thanks! Once the CI passes I'll merge.

Issue #29, opened 2016-09-03
Implementing Iterator for `Vector`


AtheMathmo:
Similarly to `Matrix` we should implement `iter` and `iter_mut` methods directly on `Vector`. This should simply call `self.data.iter()`. We can then implement `IntoIterator` using these methods.

PR #34, opened 2016-09-10
vector iteration (in the matter of #29)

zackmdavis:
[hello from Portland](http://www.meetup.com/PDXRust/events/233411353/)


AtheMathmo:
Very happy to see another contribution from you! Thanks for picking this up.

Hope you're having fun in Portland! That meetup looks like it would have been a lot of fun.

Issue #41, opened 2016-09-16
`solve_l_triangular` too harsh?


tokahuke:
My program has been panicking when I invoke `solve_l_triangular` on this:

```
⎡                    1.68933665291405 0.0000000000000000027626478732696488⎤
⎣                  0.0210783862932372                   1.7853800754297995⎦

```
Ok, I know! This is technically not lower triangular. But could not `rulinalg` be a bit more forgiving? Besides, there is no way to tell the code "I know what I am doing. Carry on!". 

Alternatively, there is no easy way to force a matrix to be lower triangular, is there?

AtheMathmo:
Yes you are probably right on this one - though I'm not too sure what the tactic here is for other libraries.

I think at the very least we should make this a failure rather than a panic - so this would be captured by the `Result` instead of terminating the program. It also seems sensible to me to use some tolerance and check that `x < eps` rather than `x == T::zero()`. Picking that `eps` value could be a little tricky though.

There is no trivial way to force a matrix to be lower triangular, but you could do something like the following:

```
for (idx, row) in a.iter_rows_mut().enumerate() {
    for x in row[idx+1..].iter_mut() {
        *x = 0.0;
    }
}
```

I haven't tested this for oversights but hopefully the intent is clear :)

Andlon:
A bit tangential, but highly related. I don't know if it has been discussed before, but would it make any sense at all to solve these constraints on the type level? I.e., you'd have types `LowerTriangularSlice`, `UpperTriangularSlice`, `DiagonalSlice` etc, and you can extract them from a general matrix using appropriate methods (i.e. `my_matrix.lower_triangular_slice()` or similar). It opens up for extra optimizations, higher degree of compile-time verification and potentially less error handling on the part of the user. That said, it might be a can of worms as the number of different types required may explode, and it may certainly complicate implementation.

With regards to this particular issue, it would avoid the problem altogether. I.e:
```
// The user explicitly tells us to ignore any elements that do not 
// belong to the lower triangular part of the matrix, 
// whether they are zero or not
let A: Matrix<T> = general_matrix_that_is_actually_lower_triangular();
let x = A.lower_triangular_slice().solve(b); 

```
Note that what I'm proposing is not to abstract away the storage of this matrix. The `.solve()` function would internally just work with `Matrix<T>` types as it currently does. This would presumably purely be a change in the API, although it can be used to provide more optimal implementations in certain cases.

[...]

PR #42, opened 2016-09-17
More lenient error handling on triangular solvers

AtheMathmo:
This resolves issue #41

### DIFF
--- a/src/matrix/slice.rs
+++ b/src/matrix/slice.rs
@@ -45,6 +45,11 @@ pub trait BaseMatrix<T>: Sized {
     /// Row stride in the matrix.
     fn row_stride(&self) -> usize;
 
+    /// Returns true if the matrix contais no elements
+    fn is_empty(&self) -> bool {
+        self.rows() == 0 || self.cols() == 0
+    }
+
     /// Top left index of the matrix.
     fn as_ptr(&self) -> *const T;
 
@@ -625,6 +630,9 @@ pub trait BaseMatrix<T>: Sized {
     /// Given a matrix `U`, which is upper triangular, and a vector `y`, this function returns `x`
     /// such that `Ux = y`.
     ///
+    /// Note that this function will not check that `U` is upper triangular but will instead use only
+    /// the upper triangular portion of the matrix.
+    ///
     /// # Examples
     ///
     /// ```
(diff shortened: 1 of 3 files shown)


src/matrix/slice.rs
```diff
@@ -625,6 +630,9 @@ pub trait BaseMatrix<T>: Sized {
     /// Given a matrix `U`, which is upper triangular, and a vector `y`, this function returns `x`
     /// such that `Ux = y`.
     ///
+    /// Note that this function will not check that `U` is upper triangular but will instead use only
+    /// the upper triangular portion of the matrix.
```
Andlon: This is a subtle thing, but I think the way it is currently phrased, it sounds as if the API expects `self` to be upper triangular, and it simply doesn't check it, which sounds like a somewhat unsafe procedure.

I think it would be helpful to make more clear the semantics of the operation: you're not assuming that `self` is a triangular matrix and not checking it. Rather, you're explicitly only concerned with the upper triangular _part_ of `self`. Hence, you could perhaps rephrase it as something 

> Given a matrix `A` and a vector `y`, this function returns the solution of the upper triangular system `Ux = y`, where `U` is the upper triangular part of `A`.

Also, out of curiosity: why `y`? The conventions I'm used to dictate that `x`, `y`, `z` are normally reserved for variable quantities. Here it's simply the constant right-hand side of a system. (sorry, don't mean to be super-nitpicky)

AtheMathmo: I was just revisiting this. When I first read the comment I agreed with the proposed doc change but now I'm not so sure.

This function is primarily for the case where `A` is an upper triangular matrix, we just happen not to check this fact (without any safety issues). I worry that your proposed change will make this primary usage a little less obvious.

> Also, out of curiosity: why `y`?

I'm not sure the convention I was taught for these expressions - if you have a more accepted alternative I'll be happy to change it!

Andlon: > I was just revisiting this. When I first read the comment I agreed with the proposed doc change but now I'm not so sure.

> This function is primarily for the case where A is an upper triangular matrix, we just happen not to check this fact (without any safety issues). I worry that your proposed change will make this primary usage a little less obvious.

I see your point. My suggested documentation change would make it more "correct", in some sense, but it would perhaps make the main use case less obvious. I suggest you _either_:

- Use your original documentation, and simply state that we are not checking the invariant.
- Keep my suggestions, but in addition emphasize the primary use case in some way.

I guess both cases are somewhat suboptimal, but they are both OK as a stopgap solution, until we can decide on a way to enforce the invariants in the type system (assuming that this is the desirable ultimate solution, which I think it is?).

AtheMathmo:
@Andlon I went with your doc change in the end - it did make the most sense :)

Andlon:
@AtheMathmo: all right! Do you feel ready to merge, or what do you feel is missing?

AtheMathmo:
If you're happy with it I'm ready to merge!

Andlon:
Yeah, I think it's good to go. It's not 100% optimal, but it's a better solution than what is presently in `rulinalg`, and a better solution will take some time to formulate/agree on (as it pertains to the more general question of how to deal with matrix constraints).

PR #43, opened 2016-09-17
Added convenience constructor function for Matrix.

regexident:


AtheMathmo:
Thanks for opening this PR!

I think this makes a neat addition and the implementation looks good. There is one minor point here - in the `new` function we allow empty matrices to be constructed. I see no issue with allowing this here as well. If we want to keep panicking then we should make this clear in the documentation. We should let the user know that `rows`/`cols` must both be non-zero. This should be in a [`Panics` section](https://doc.rust-lang.org/book/documentation.html#special-sections).

Also I'd like to make sure the use case is justifies the addition of this function - keeping the API lean is important I think (even if it's not done too well so far). Could you share a little why you need this function or think it should be added?

regexident:
> I think this makes a neat addition and the implementation looks good. There is one minor point here - in the new function we allow empty matrices to be constructed. I see no issue with allowing this here as well. If we want to keep panicking then we should make this clear in the documentation. We should let the user know that rows/cols must both be non-zero. This should be in a Panics section.

Good point. I removed the assert.

> Also I'd like to make sure the use case is justifies the addition of this function - keeping the API lean is important I think (even if it's not done too well so far). Could you share a little why you need this function or think it should be added?

And another good point. Mea culpa. The example should now serve as a motivation. 

Basically  I find myself frequently needing to construct a matrix by calculating its values from external data, most notably in the form of distance matrices. The `Matrix::by` aims to provide a convenience constructor for these scenarios while encapsulating the matrix' internal representation (row-major, column-major?).

AtheMathmo:
Thanks for giving some motivation - I agree that the new example shows a good use case.

I think the only thing missing here is an explanation of how the function works in the documentation. Although the example makes it clear I think it's best to state explicitly so that the user doesn't have to figure it out.

Maybe we can just add something like:

_Takes a function `f` and constructs a new matrix such that A<sub>i j</sub> = f(i, j)._

regexident:
> Maybe we can just add something like:
> 
> Takes a function f and constructs a new matrix such that Ai j = f(i, j), where i is the row index and j the column index.

[Fixed.](https://github.com/regexident/rulinalg/blob/2b366b0067c363e10d51488a6b7ec534253ae2e2/src/matrix/mod.rs#L129-L131)

theotherphil:
Bikeshedding, but I find the name of this function unintuitive. The image crate calls the equivalent function  from_fn. Could we use something like that here?

theotherphil:
+1 on adding the function. Another use case for it is generating synthetic data sets for tests.

regexident:
> Bikeshedding, but I find the name of this function unintuitive. The image crate calls the equivalent function from_fn. Could we use something like that here?

I'm not emotionally attached to `fn by`. I chose it as `_by` is a common function name suffix in the stdlib for function variants that allow for a closure to be provided. How about `fn new_by`?

> +1 on adding the function. Another use case for it is generating synthetic data sets for tests.

Yes, indeed!

Andlon:
I think this is a most welcome addition!

Naming might feel like bikeshedding, but I think intuitive naming is extremely important in order to ensure a good user experience. I think maybe `from_fn` has been the best suggestion so far. Particularly because typing `Matrix::from` will give you available constructors when autocompleting code. I've observed that a `from` prefix seems to be idiomatic for constructors, whereas `by` seems to be used primarily to operate on existing objects, such as `by_ref()` or `sort_by()`.

AtheMathmo:
Thanks for taking the time to read this! I agree with other's comments here, `from_fn` feels the most natural.

regexident:
So be it then! Changed the name to `fn from_fn`.

Thanks for the great feedback, all!

src/matrix/mod.rs
```diff
@@ -769,6 +807,19 @@ mod tests {
     }
 
     #[test]
+    fn test_mat_by() {
```
AtheMathmo: We should also rename the test so it is easy to find and maintain in future

AtheMathmo:
One more very minor change then I'm happy to merge this. Anything else from @Andlon or @theotherphil ?

src/matrix/mod.rs
regexident: Absolutely. Missed it. Fixed now.

src/matrix/mod.rs
```diff
@@ -126,6 +126,44 @@ impl<T> Matrix<T> {
         }
     }
 
+    /// Constructor for Matrix struct that takes a function `f`
+    /// and constructs a new matrix such that `Ai j = f(i, j)`,
```
Andlon: `Ai j` looks like there is an extra space. So perhaps `Aij` or `A_ij` or so? Sorry for the nitpick, I just noticed it and I can not un-see it!

AtheMathmo: Actually yes I agree with this. Maybe when the doc renders it will look fine - but maybe not. I'd change to `A_ij` - that should be universally clear?

Andlon: I was curious, so I actually rendered it to check, and it looks like on Github. Just to clear up confusion.

regexident: Changed it to `A_ij`. Thanks Andlon!

Had copied the sentence from AtheMathmo's GitHub comment, which added a space in there for whatever reason. ¯\\\_(ツ)\_/¯

AtheMathmo:
Thanks @regexident ! Will merge this now.

Issue #40, opened 2016-09-14
Custom types for decompositions


Andlon:
Rulinalg has implementations of a number of decompositions. Currently these are returned as `Result<tuple, Error>`, where `tuple` is a tuple of `Matrix<T>`. This is in itself useful, as it gives the user easy access to the decomposed parts. However, often you are not necessarily so interested in the decomposed parts themselves, but rather what you can do with them.

For example, in the case of LU or QR and a linear system `Ax = b`, you may want to perform the decomposition once, and then use it many times over to solve for different instances of `b`. In this case, as a user, you don't want to have to perform the transposition of Q and the solution of the triangular system. Rather, it would be convenient if you could do something along the lines of:

    let qr = A.qr();
    for b in right_hand_sides {
        let x = qr.solve(b);
        do_something(x);
    }

This is possible if `qr()` returns a struct `QR` that has a specialized `solve` method, as well as accessors to the decomposed parts `Q` and `R`. Similarly for `LU`. In my opinion, you would not lose anything over the current API, but you'd make it easier to use, and it could also open up for optimizations in how the decompositions are stored, and how they are used to perform certain actions.

For SVD, you could have convenient methods to retrieve the singular values, perhaps for least squares solution etc. In addition to accessors to the decomposed parts, of course.

Eigen (C++ library) [does something like this](https://eigen.tuxfamily.org/dox/group__TutorialLinearAlgebra.html). I'll be quick to add that I do not want `rulinalg` to mirror Eigen's API too closely, as it is... perhaps overly complex at times.

Do you think this style of API could be suitable for `rulinalg`? I'd be interested to hear your opinions!

AtheMathmo:
Thanks for writing this up -I think it's a great idea!

It certainly makes the decompositions easier to use in some cases and makes the signatures of some functions nicer. I worry that the api could be a little too verbose though, for example I could see things like the following being inevitable:

```rust
let (q, r) = A.qr().unwrap().get_parts();
```

I still think that overall it is a lot cleaner - but documentation will be essential to keep things obvious to the end user.

I also wonder if there is a way to use this to help us access the different types of algorithms for each decomposition. For example for SVD we may only want the singular values which takes less time. Or even eigenvalues/vector algorithms. Right now the only way we can do this is by providing different functions - but perhaps there is a tidier way?

I'm not sure I'll have time to work on this for a while but I'd be really happy if someone else had the time to pick it up.

Andlon:
[...]

PR #44, opened 2016-09-18
[POC] Creating a new wrapper struct for eigen methods

AtheMathmo:
Per discussion in #40 I wanted to see what a new custom type for eigen decomposition would look like. This adds an intermediate between computing the eigen values and eigen vectors for a matrix. The motivation for doing so is to make the API cleaner - though there are more benefits in this approach for other algorithms.

## The change

Previously to compute eigen values we would do the following:

```rust
// `A` is a matrix
let values = A.eigenvalues().unwrap();
```

Now this will look like this:

```rust
// `A` is a matrix
let values = A.eigen().values().unwrap();
```

The `eigen` invocation will do no actual work but will take ownership of `A` inside an `Eigen` struct. We can then invoke `values` or `decomp`.

## Advantages

- We can create new eigen-thing algorithms and they will be easier for our users to find.
- The code is better organized (though we can achieve this without the `Eigen` abstraction).
- We now have a place to implement LAPACK routines and can even provide direct access to them via the `Eigen` struct.

## Disadvantages

- Is this unnecessarily complicated?

There is also one especially untidy part of this story which is amplified by this change. For the general Francis shift implementations we take ownership now as otherwise we invoke a `clone`. However we do not need to take ownership for the 2x2 case - in order to access this case for the svd algorithm I have left the 2x2 eigenvalue function in the `decomposition` module.

If anyone has any better suggestions for handling this I'd appreciate it.

### DIFF
--- a/src/matrix/decomposition/mod.rs
+++ b/src/matrix/decomposition/mod.rs
@@ -10,11 +10,14 @@
 //! 3. [Computation of the SVD]
 //! (http://www.cs.utexas.edu/users/inderjit/public_papers/HLA_SVD.pdf)
 
+pub mod eigen;
+
 use std::any::Any;
 use std::cmp;
 use std::ops::{Mul, Add, Div, Sub, Neg};
 use std::slice;
 
+use self::eigen::Eigen;
 use matrix::{Matrix, MatrixSlice, MatrixSliceMut, BaseMatrix, BaseMatrixMut};
 use vector::Vector;
 use Metric;
@@ -22,7 +25,7 @@ use utils;
 use error::{Error, ErrorKind};
 
 use libnum::{One, Zero, Float, Signed};
-use libnum::{cast, abs};
+use libnum::cast;
 
 impl<T: Any + Float> Matrix<T> {
     /// Cholesky decomposition
@@ -434,7 +437,7 @@ impl<T: Any + Float + Signed> Matrix<T> {
             }
         }
 
-        let c_eigs = try!(c.eigenvalues());
+        let c_eigs = try!(c.direct_2_by_2_eigenvalues());
 
         // Choose eigenvalue closes to c[1,1].
         let lambda: T;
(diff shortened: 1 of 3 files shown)


Andlon:
Currently `.eigenvalues()` and `.eigendecomp()` only take `&self`, so with the proposed solution, you'd have to copy your matrix if you wanted to use it after computing these quantities.

Andlon:
Sorry, I hadn't looked at the code in detail. That's because it internally called clone earlier, but now it doesn't anymore. Then please disregard my previous comment!

src/matrix/decomposition/mod.rs
```diff
@@ -10,19 +10,22 @@
 //! 3. [Computation of the SVD]
 //! (http://www.cs.utexas.edu/users/inderjit/public_papers/HLA_SVD.pdf)
 
+pub mod eigen;
```
Andlon: Does the eigen module need to be exposed to the user? It's useful for internally organizing things, but I think for the user it might perhaps just be an obstacle?

src/matrix/mod.rs
```diff
@@ -16,7 +16,7 @@ use error::{Error, ErrorKind};
 use utils;
 use vector::Vector;
 
-mod decomposition;
+pub mod decomposition;
```
Andlon: Same as with the comment about the `eigen` module. What are the benefits of exposing this to the public?

Andlon:
From a brief look, I think it looks good. I don't fully understand the benefits of making more modules public. I made a couple of short comments on that.

Not directly related to the question at hand (wrapper struct), but I think it's not possible to determine which eigenvectors are "true" and which ones are generalized with the current application, is that right? Would users need to know this, perhaps? I have no idea, it just came to mind.

src/matrix/mod.rs
AtheMathmo: This might not be necessary actually. I exposed them because I wasn't getting errors when the public functions on `Eigen` weren't documented. Maybe these are still accessible even if the mods are private. There is also some documentation in the `eigen` module that I wanted to be accessible.

Andlon: I must admit I still don't grasp Rust's module system very well. If you re-export everything from the `eigen` module into the main module (`matrix`?), does that still give all the issues you mentioned?

AtheMathmo: I guess that as the documentation is my primary concern I should just check how it works. I'll get back to you soon!

Andlon:
@AtheMathmo: for the problem of duplicated 2x2 eigenvalues, perhaps you could just move it into a separate, internal module, from which both `eigen` and `svd` could use?

AtheMathmo:
I think this will be the necessary approach. I'll probably have to keep the 2x2 eigenvalues in the `decomposition` module (as it is now) and move the others out into sub-modules so they can use the private function.

src/matrix/mod.rs
AtheMathmo: It looks like if the module isn't made public then rustdoc will not provide any API information on the `Eigen` struct (unless it is reexported into a public module - I think). Because of this I think it is worth making the `decomposition` and `eigen` modules public. We'll also need to update the decomposition module docs.

AtheMathmo:
Closing due to #66

PR #50, opened 2016-09-24
Added impl for From/Into<Vec<T>> to Vector<T>.

regexident:
**Motivation**:
Basically the common conveniences associated with `From` and `Into` traits, such as being able to pass `vec![…]` to `fn foo(vec: Vector<T>) { … }` instead of `Vector::new(vec![…])`.

Also it just kinda makes sense being a shallow wrapper around `Vec<T>`.

### DIFF
--- a/src/convert.rs
+++ b/src/convert.rs
@@ -8,6 +8,17 @@ use std::convert::From;
 use super::matrix::{Matrix, MatrixSlice, MatrixSliceMut, BaseMatrix};
 use super::vector::Vector;
 
+impl<T> From<Vec<T>> for Vector<T> {
+    fn from(vec: Vec<T>) -> Self {
+        Vector::new(vec)
+    }
+}
+
+impl<'a, T> From<&'a [T]> for Vector<T> where T: Clone {
+    fn from(slice: &'a [T]) -> Self {
+        Vector::new(slice.to_owned())
+    }
+}
 
 impl<T> From<Vector<T>> for Matrix<T> {
     fn from(vector: Vector<T>) -> Self {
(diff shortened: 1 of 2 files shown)


AtheMathmo:
This is another really great addition, thank you!

I won't have time to check over the code until tomorrow. Also do you know if this will interact with operator overrides at all? For example can we add `Vec`s like this?

AtheMathmo:
I think that we should move this into the `convert` module. Otherwise this is good to merge.

_Looks like we can't coerce `Vec` addition like this - unless I missed something._

regexident:
> I think that we should move this into the convert module. Otherwise this is good to merge.

I've moved the impls for `From<…>` into `convert.rs`. The impl for `Into<…>` however needs access to `self.data`, which is private, and thus cannot be moved out of `vector.rs`.

> Looks like we can't coerce Vec addition like this - unless I missed something.

Well, instead of …

```rust
impl<'a, T: Copy + Add<T, Output = T>> Add<Vector<T>> for &'a Vector<T> {
    type Output = Vector<T>;

    fn add(self, v: Vector<T>) -> Vector<T> {
        v + self
    }
}
```

… one would have to use something like this …

```rust
impl<'a, T, V> Add<V> for &'a Vector<T>
    where T: Copy + Add<T, Output = T>,
          V: Into<Vector<T>> {
    type Output = Vector<T>;

    fn add(self, v: V) -> Vector<T> {
        let v: Vector<T> = v.into();
        v + self
    }
}
```

AtheMathmo:
That's great, thanks!

As for the adding, I was hoping we could do:

```rust
let b = vec_a.into() + vec_b.into()
```

But I couldn't get that to compile. It's not a big deal. I think I would prefer to keep the arithmetic as `Vector`  to `Vector` right now - for no good reason other than fear of the unknown. But we should definitely make use of this in other areas of the library!

PR #51, opened 2016-09-25
Sparse matrix (COO, CSC, CSR) implementations

c410-f3r:
This initial pull request is intended to discussion the viability of sparse matrix implementations and code review


AtheMathmo:
Thanks for getting this PR in so quickly! I've only had a very brief look through, here are some minor comments:

- This is a little inconsistent with existing method names, and idiomatic rust. For example generally you don't see `get_field` functions, instead it would just be called `field`. You should also check the existing library functions to aim to be as consistent as possible, `get_rows` -> `rows` etc.
- If you could have a think about how we can crawl towards implementing the `BaseMatrix` trait this would be a good idea. We won't want it for this PR though as it requires slice objects and some other trickery.
- It's awesome to see so many unit tests - if you could add some basic benchmarks I would really appreciate it too!
- Some nitpicks on docs. I'm pretty picky on these so we'll address as we go through review. But one big point is that the examples you've added to the `SparseMatrix` trait implementations wont be seen in docs (which would be a shame as they are good!). Instead these examples should be in the trait definition. You can only add visible docs per implementation at the top of the `impl` block, i.e.

```rust
/// Implements Trait for Thing
///
/// And these docs will be rendered
impl Trait for Thing {
    /// This wont be rendered in docs
    /// but example tests here should still run.
    fn foo(&mut self);
}
```

All in all I think that the spec and layout are really solid. I can't really comment on efficiency etc. as it's not my area of expertise but this doesn't need to be perfect off the bat (the rest of the library certainly isn't).

I think you have a good amount of features in here for an initial merge. We may want to add some more arithmetic to the mix but I'd even be happy forgoing this for now (just to make the merges/testing a little more manageable). Were you planning on including anything else before merging?

Thanks for your hard work on this!

AtheMathmo:
Oh and one more request. If you have any sources you've been using for this - whether online or otherwise I think it would be valuable to reference them in the relevant module docs. I've done this in a few places (like the `decomposition` module). It will make it easier for someone else to pick up later.

src/sparse_matrix/compressed_matrix/csc_matrix/mod.rs
```diff
@@ -0,0 +1,280 @@
...
+            rows: rows,
+            cols: cols,
+            indices: indices,
+            nnz: nnz,
+            ptrs: ptrs,
+            values: values,
+        }
+    }
+
+    fn get_indices(&self) -> &Vec<usize> {
```
AtheMathmo: Let's have these return slices (&[usize])

src/sparse_matrix/compressed_matrix/csc_matrix/mod.rs
```diff
@@ -0,0 +1,280 @@
...
+        &self.indices
+    }
+    fn get_ptrs(&self) -> &Vec<usize> {
+        &self.ptrs
+    }
+    fn get_values(&self) -> &Vec<T> {
+        &self.values
+    }
+    fn set_indices(&mut self, indices: &Vec<usize>) {
```
AtheMathmo: And again the input arguments should be slices

src/sparse_matrix/compressed_matrix/csc_matrix/mod.rs
```diff
@@ -0,0 +1,280 @@
...
+    fn to_csr(&self) -> CsrMatrix<T> {
+        CsrMatrix::new(self.rows,
+                       self.cols,
+                       self.indices.clone(),
+                       self.ptrs.clone(),
+                       self.values.clone())
+    }
+}
+
+impl<T: Clone> Clone for CscMatrix<T> {
```
AtheMathmo: We should be able to derive this instead of explicitly implementing (and for the others too)

src/sparse_matrix/compressed_matrix/csr_matrix/csr_matrix_arithm.rs
```diff
@@ -0,0 +1,25 @@
...
+impl<T: Copy + Mul<T, Output = T>> Mul<T> for CsrMatrix<T> {
+    type Output = CsrMatrix<T>;
+
+    fn mul(self, scalar: T) -> CsrMatrix<T> {
+        self * &scalar
+    }
+}
+
+/// Multiplies CSR matrix by scalar.
+impl<'a, T: Copy + Mul<T, Output = T>> Mul<&'a T> for CsrMatrix<T> {
```
AtheMathmo: We also want to implement these `for &'b CsrMatrix`.

AtheMathmo:
And one last point for now: if I find myself _needing_ to clone the object for a function I tend to prefer to let that function consume the object - then the user can choose to clone if they need to. I think this is the case in a few of your functions.

c410-f3r:
Thank you very much for reviewing the code! I am going to modify everything according to your feedback and send another PR when I implement most of the Matrix/BaseMatrix methods (det, variance, etc).

AtheMathmo:
I think you understood but to be clear: I think you should try to keep it minimal for this PR. The functionality you have is a great start - I think only the arithmetic needs to be fleshed out a little (matrix x vector, matrix x matrix). With the `BaseMatrix` comment I simply meant to try to think whether your design could easily fit into the functions needed. The other stuff can come in a later PR :)

Andlon:
Hey @c410-f3r, good job! So far I've only had time to skim through quickly, and I have a generally good impression. I need some more time to give proper feedback though. I'll try to come back to you during the next few days. @AtheMathmo is pretty spot on with his comments.

For now, I have the following high-level feedback:

- Perhaps we can reduce the number of public modules? I.e., from a user's perspective, I'd ideally like to just be able to do `use rulinalg::sparse::CsrMatrix`. You can use several internal modules to keep things organized, but we can hide this from the user for a more ergonomic usage experience.
- Do we actually need an explicit `CooMatrix`? I believe COOMatrix in SciPy exists primarily because i.e. lists in Python would be too slow. However, in Rust, we could use generic programming to construct e.g. a `CsrMatrix` from any `[Coordinate]`, where `Coordinate` is a trait that exposes the (i, j, v) coordinate triplet. So, perhaps, `fn from_coordinates<T>(coords: &[T]) where T: Coordinate` or so. In fact, it may be possible to construct a sparse matrix from an iterator of coordinates, which could potentially let the user build the matrix lazily without having to allocate a huge buffer of coordinate triplets. I'm not saying either is right or wrong here, but it's something we should discuss. However, having an explicit `CooMatrix` for now is certainly acceptible and useful in its current state!
- If we keep `CooMatrix`, perhaps using `Into/From` traits instead of `to_csr()` and similar explicit methods?

Overall, good work, and I'll try to come back with a more detailed technical feedback soon!

src/sparse_matrix/mod.rs
```diff
@@ -0,0 +1,45 @@
...
+use sparse_matrix::coordinate::Coordinate;
+use sparse_matrix::compressed_matrix::csc_matrix::CscMatrix;
+use sparse_matrix::compressed_matrix::csr_matrix::CsrMatrix;
+
+/// Contract for sparse matrices implementation
+pub trait SparseMatrix<T> {
+    /// Constructs matrix with given coordinates (rows, cols and values).
+    ///
+    /// Requires slice of coordinates.
+    fn from_coordinates<C>(coords: &[C]) -> Self where C: Coordinate<T>;
```
AtheMathmo: I think it would be better here to be generic over objects implementing `IntoIterator`, without testing I think a signature like:

```rust
fn from_coordinates<U, C>(coords: U) -> Self
    where U: IntoIterator<Item=C>, C: Coordinate<T>;
```

This will let us use slices still - or `Vec`s etc. If you need more than an iterator to do the construction we can keep it as-is. Note also you can get the length of the iterator by requiring [`ExactSizeIterator`](https://doc.rust-lang.org/std/iter/trait.ExactSizeIterator.html) too.

src/sparse_matrix/mod.rs
```diff
@@ -0,0 +1,45 @@
...
+    ///
+    /// Requires slice of diagonal elements.
+    fn from_diag(diag: &[T]) -> Self;
+    /// Constructs the identity matrix.
+    ///
+    /// Requires the size of the matrix.
+    fn identity(size: usize) -> Self;
+
+    /// Returns number of rows
+    fn rows(&self) -> usize;
```
AtheMathmo: These are perfect here for now. Later we will _hopefully_ be able to do `pub trait SparseMatrix: BaseMatrix` and implement some of these functions there.

src/sparse_matrix/compressed_matrix/mod.rs
```diff
@@ -0,0 +1,33 @@
...
+           -> Self;
+
+    /// Returns indices
+    fn indices(&self) -> &[usize];
+    /// Returns pointers (offsets)
+    fn ptrs(&self) -> &[usize];
+    /// Returns values
+    fn values(&self) -> &[T];
+    /// Sets indices
+    fn set_indices(&mut self, indices: &Vec<usize>);
```
AtheMathmo: The comment below was before I had seen another piece of code. I've left it in case it is worth reading.

---

Also - we should have a think about whether we really want to expose this functionality. I have a feeling that we probably don't.

On the matrix side we never let the user modify the **shape** of the `Matrix` as this would lead to dramatic breakage in functions. We need to guarantee that `self.rows * self.cols == self.data.len()`. If we allow users to modify the `values`, `indices`, and `ptrs` wont they be able to break similar requirements in the sparse matrices?

I think a better alternative is to have these function return mutable slices:

```rust
fn mut_indices(&mut self) -> &mut [usize];

fn mut_ptrs(&mut self) -> &mut [usize];

fn mut_values(&mut self) -> &mut [usize];
```

This way the user can modify values in place but cannot change the size of these after the matrix has been constructed. I'll have to defer to you and others to determine whether this still leaves some safety issues - for example can the user modify the pointers to point to invalid data?

src/sparse_matrix/compressed_matrix/compressed_matrix_utils.rs
```diff
@@ -0,0 +1,58 @@
+use libnum::{One, Zero};
+
+use sparse_matrix::compressed_matrix::CompressedMatrix;
+
+fn expand_ptrs_indices_values<C>(ptrs: &[usize], mut closure: C)
+    where C: FnMut(usize, usize)
+{
+    for idx in (0..(ptrs.len() - 1)).rev() {
+        let greater_ptr = ptrs[idx + 1];
```
AtheMathmo: To remove bound checks here I think we can do the following:

```rust
for (lesser_ptr, greater_ptr) in ptrs.windows(2).rev() { ... }
```

I believe that should give the same functionality.

src/sparse_matrix/compressed_matrix/csc_matrix/mod.rs
```diff
@@ -0,0 +1,258 @@
...
+    }
+    fn ptrs(&self) -> &[usize] {
+        &self.ptrs.as_slice()
+    }
+    fn values(&self) -> &[T] {
+        &self.values.as_slice()
+    }
+    fn set_indices(&mut self, indices: &Vec<usize>) {
+        if indices.len() == self.nnz {
+            self.indices = indices.clone();
```
AtheMathmo: I think this may invalidate some of my other comments.

As we are cloning here I think it is better to take a `Vec<usize>` instead of a reference to one for the indices argument. If we need to take ownership of the object it is better to reflect this in the signature and the user can decide if they need to `clone`.

This is the same in the two other functions below.

src/sparse_matrix/compressed_matrix/csc_matrix/mod.rs
```diff
@@ -0,0 +1,258 @@
...
+    /// ```
+    /// use rulinalg::sparse_matrix::SparseMatrix;
+    /// use rulinalg::sparse_matrix::compressed_matrix::CompressedMatrix;
+    /// use rulinalg::sparse_matrix::compressed_matrix::csc_matrix::CscMatrix;
+    ///
+    /// let _ = CscMatrix::new(4, 4, vec![0, 0, 0], vec![0, 1, 2, 3, 3], vec![1, 2, 3]).transpose();
+    /// ```
+    ///
+    /// Complexity: O(rows + max(cols, nnz))
+    fn transpose(&self) -> CscMatrix<T>
```
AtheMathmo: As mentioned before this function should just consume `self` if we need to do a `clone` anyway. Give the user the choice :)

src/sparse_matrix/compressed_matrix/csc_matrix/mod.rs
```diff
@@ -0,0 +1,258 @@
...
+    ///
+    /// ```
+    /// use rulinalg::sparse_matrix::SparseMatrix;
+    /// use rulinalg::sparse_matrix::compressed_matrix::CompressedMatrix;
+    /// use rulinalg::sparse_matrix::compressed_matrix::csc_matrix::CscMatrix;
+    ///
+    /// let csc_mat = CscMatrix::new(4, 4, vec![0, 0, 0], vec![0, 1, 2, 3, 3], vec![1, 2, 3]).to_csc();
+    /// ```
+    fn to_csc(&self) -> CscMatrix<T> {
+        self.clone()
```
AtheMathmo: The `clone` here is acceptable as generally the `to_` format warns the user that this operation may be expensive.

We may revisit whether we want to use these functions or `Into/From` later though.

src/sparse_matrix/coordinate.rs
```diff
@@ -0,0 +1,11 @@
+//! Everything related to coordinate
+
+/// Represents a coordinate (row and column) with repective value
+pub trait Coordinate<T> {
```
AtheMathmo: I think it is a good idea for us to provide a default implementation of `Coordinate` for triplets.

`impl<T: Copy> Coordinate<T> for (usize, usize, T) { ... }`

c410-f3r:
Thanks @AtheMathmo  and @Andlon ! Soon I will make more changes

src/sparse_matrix/compressed_matrix/csc_matrix/mod.rs
```diff
@@ -0,0 +1,258 @@
...
+    use sparse_matrix::compressed_matrix::CompressedMatrix;
+    use sparse_matrix::compressed_matrix::csc_matrix::CscMatrix;
+    use sparse_matrix::compressed_matrix::csr_matrix::CsrMatrix;
+    use sparse_matrix::SparseMatrix;
+
+    #[test]
+    fn test_equality() {
+        let a = CscMatrix::new(3, 3, vec![0, 1, 1, 2], vec![0, 1, 2, 2], vec![1, 2, 3, 4]);
+        let b = a.clone();
+        assert_eq!(a, b);
```
Andlon: Also need a test for something that is not equal here, in case some devil of a person decides to make your equality routine always return `true` ;)

src/sparse_matrix/compressed_matrix/csc_matrix/mod.rs
```diff
@@ -0,0 +1,258 @@
...
+    }
+
+    fn rows(&self) -> usize {
+        self.rows
+    }
+    fn cols(&self) -> usize {
+        self.cols
+    }
+    fn nnz(&self) -> usize {
+        self.nnz
```
Andlon: I'm a little tired now, so maybe I'm missing something obvious, but isn't `nnz` simply equal to `values.len()`? (in which case we could trim off a redundant member in our matrix)

src/sparse_matrix/compressed_matrix/mod.rs
Andlon: I agree that this pattern might easily lead to bugs. Another way could perhaps be to remove these functions and only let the constructor remain (which consumes its input), and then additionally provide a function that consumes the entire matrix and gives you the raw parts, which can be fed into a new matrix. Some users might actually want access to this if they're implementing their own algorithms on top of our sparse data structures.

src/sparse_matrix/mod.rs
Andlon: I'm by no means an expert on the implementation of these data structures (my experience with sparse linear algebra comes primarily from library usage), but I believe most implementations (except for DOK matrices) must usually be sorted in some way, so one might even have to take ownership of a vec or something.

Moreover, I'm not sure if this particular function should be in the `SparseMatrix` trait, as it may not be natural to do it for every matrix. For example, I mentioned DOK (dictionary of keys, basically just a hashmap with coordinates as keys) matrices, in which case you'd just access elements directly, and I don't think you'd ever want to construct a DOK matrix from coordinate triplets.

src/sparse_matrix/compressed_matrix/mod.rs
AtheMathmo: This sounds like a sensible approach! It will be a little more verbose for the user but _much_ safer and easier for us to handle user-error.

src/sparse_matrix/compressed_matrix/csc_matrix/mod.rs
c410-f3r: Indeed, nnz is equal to `values.len()` and `indices.len()`. It is a guarantee that `values` and `indices` have the given size and is a central reference for quick access.

src/sparse_matrix/compressed_matrix/compressed_matrix_utils.rs
```diff
@@ -0,0 +1,53 @@
...
+        let lesser_ptr = ptrs[idx];
+        for value in (lesser_ptr..greater_ptr).rev() {
+            closure(idx, value)
+        }
+    }
+}
+
+pub fn transpose<M, T>(compressed_matrix: &M, ptrs_size: usize) -> (Vec<usize>, Vec<usize>, Vec<T>)
```
AtheMathmo: I just had a thought - as we have support for both CSC and CSR matrices we can get a free transpose - we simply swap the dimensions and the matrix type (while using the same underlying data). We should definitely support this - whether it is the default or not I'm not sure. I think we want to have both as possibilities though (so the user can transpose a CSR matrix and keep it as CSR).

c410-f3r: Indeed, Scipy itself transposes CSR returning a new instance of CSC.
For now, I will implement transpose_csc and transpose_csr until we find a better approach :p

src/sparse_matrix/compressed_matrix/csc_matrix/mod.rs
```diff
@@ -40,6 +40,29 @@ impl<T: Copy + One + Zero> CscMatrix<T> {
...
+    /// let _ = CscMatrix::new(4, 4, vec![0, 0, 0], vec![0, 1, 2, 3, 3], vec![1, 2, 3]).transpose();
+    /// ```
+    ///
+    /// Complexity: O(1)
+    pub fn transpose_csr(&self) -> CsrMatrix<T>
+        where T: Copy
+    {
+        CsrMatrix::new(self.rows,
+                       self.cols,
+                       self.indices.clone(),
```
AtheMathmo: This transformation is **not** free as we're cloning the data here. To make it free we should consume `self` and just pass the values directly.

c410-f3r:
A Triplet does not represent

src/sparse_matrix/compressed_matrix/compressed_matrix_utils.rs
c410-f3r: Still didn't figure out how to change the matrix struct, any clues?

AtheMathmo:
The ci is failing on these because of the `assert_ne!` macro, which is only on nightly. I guess you'll have to change to `assert!(a != b)`

src/sparse_matrix/compressed_matrix/compressed_matrix_utils.rs
AtheMathmo: I'm not sure how either without something hacks like an associated type with a marker (I.e. csc has a csr associated type). Even that might not work.

I'll see if I can figure something out tomorrow...

AtheMathmo: Sorry, I guess I'm tired. There is a way to get it working for the utils function at least.

Add a new generic type, `U: CompressedMatrix`, and give that trait a new `from_raw_parts` function that returns `Self`. Then at the end of the transpose use this to return `U`. I think that should work.

AtheMathmo:
I think that this is getting pretty close. There are still a few functional changes I think we need and maybe some slight API modifications. I'll hopefully be able to do another full review soon.

Thanks for your continued work on this @c410-f3r !

src/sparse_matrix/compressed_matrix/compressed_matrix_utils.rs
```diff
@@ -0,0 +1,98 @@
+use libnum::{One, Zero};
+
+use sparse_matrix::{CompressedMatrix, MatrixCoordinate, Triplet};
+
+fn expand_ptrs_indices_values<C>(ptrs: &[usize], mut closure: C)
```
AtheMathmo: I think it would be good to have a comment describing exactly what this function does. I'm still not too sure :)

src/sparse_matrix/compressed_matrix/csc_matrix/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+    /// # Examples
+    ///
+    /// ```
+    /// use rulinalg::sparse_matrix::{CompressedMatrix, CscMatrix, SparseMatrix};
+    ///
+    /// let _ = CscMatrix::new(4, 4, vec![0, 0, 0], vec![0, 1, 2, 3, 3], vec![1, 2, 3]).transpose();
+    /// ```
+    ///
+    /// Complexity: O(1)
+    pub fn transpose_csr(&self) -> CsrMatrix<T>
```
AtheMathmo: I still think that this function should consume `self` and not clone the indices. Especially as we advertise it as free.

src/sparse_matrix/compressed_matrix/csc_matrix/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+            rows: rows,
+            cols: cols,
+            indices: indices,
+            ptrs: ptrs,
+            values: values,
+        }
+    }
+
+    fn indices(&self) -> &[usize] {
+        &self.indices.as_slice()
```
AtheMathmo: Do we need `as_slice` here? I _think_ auto deref should do the work for us.

It doesn't matter much though!

src/sparse_matrix/compressed_matrix/csc_matrix/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+
+impl<T: Copy + One + Zero> SparseMatrix<T> for CscMatrix<T> {
+    /// # Examples
+    ///
+    /// ```
+    /// use rulinalg::sparse_matrix::{CompressedMatrix, CscMatrix, SparseMatrix};
+    ///
+    /// let csc_mat = CscMatrix::from_diag(&[1.0,2.0,3.0,4.0]);
+    /// ```
+    fn from_diag(diag: &[T]) -> CscMatrix<T> {
```
AtheMathmo: I think it would be better to take an `Into<Vec<T>>` here. Then we can simply call `into()` and get the length etc.

src/sparse_matrix/compressed_matrix/csc_matrix/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+    /// ```
+    /// use rulinalg::sparse_matrix::{CompressedMatrix, CscMatrix, SparseMatrix};
+    ///
+    /// let _ = CscMatrix::new(4, 4, vec![0, 0, 0], vec![0, 1, 2, 3, 3], vec![1, 2, 3]).transpose();
+    /// ```
+    ///
+    /// Complexity: O(rows + max(cols, nnz))
+    fn transpose(&mut self)
+    {
+    	swap(&mut self.rows, &mut self.cols);
```
AtheMathmo: Very minor point - I would prefer to `use mem` and see the full name here. I think anytime we use functions like this it is best to be explicit. It may be obvious enough but we also have `ptr::swap` in the standard library.

src/sparse_matrix/mod.rs
```diff
@@ -0,0 +1,47 @@
...
+    /// Returns number of non zero elements
+    fn nnz(&self) -> usize;
+
+    /// Tranposes the given matrix
+    fn transpose(&mut self);
+}
+
+/// Matix Coordinate (i, j)
+#[derive(Debug, Clone, Copy)]
+pub enum MatrixCoordinate {
```
AtheMathmo: I'm not too convinced that we need this enum? Perhaps I am missing some intricacy. We can keep it for now.

In any case the enum fields should be capitalized like this: `{ Row, Col }`.

src/sparse_matrix/triplet.rs
```diff
@@ -0,0 +1,33 @@
...
+    fn row(&self) -> usize {
+        self.0
+    }
+    fn col(&self) -> usize {
+        self.1
+    }
+    fn value(&self) -> T {
+        self.2
+    }
+    fn from_coordinate(&self, coo: MatrixCoordinate) -> usize {
```
AtheMathmo: A `from_` function implies that this function returns a `Triplet`. We should rename this function to `coordinate(coord: MatrixCoordinate)` - and make it clear that it returns the specified coordinate value.

I still believe we should be able to avoid this function. Perhaps with macros or in some other way.

c410-f3r:
Thanks for the notes and the patience @AtheMathmo!

AtheMathmo:
Hey @c410-f3r I just wanted to check that you're still happy to work on this? If not I'd be happy to pick up from where you are currently.

c410-f3r:
Did like 4 complete rewrites to eliminate code duplication in CSC/CSR and I finally got a reasonable solution that soon will be land. Sorry for not keeping in touch.

AtheMathmo:
No worries, thanks for the update!

PR #52, opened 2016-09-25
Return 0 from det() instead of LUP error

andrewcsmith:
LUP factorization will return an error whenever one of the diagonals of the U
matrix is == 0. This generally makes sense, because such a matrix is
singular and there will probably be an error down the line anyway. However, the
determinant of this matrix is known, and is 0, because the determinant is
calculated as the product of the diagonal of the triangular matrix.

I believe that it is more general to return the correct result for the
determinant (0) rather than raising an error. Let the end-user decide what to
do about this determinant of 0.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -9,17 +9,15 @@ use std::marker::{Send, Sync};
 /// An error related to the linalg module.
 #[derive(Debug)]
 pub struct Error {
-    /// Type of error that led to this issue
-    pub kind: ErrorKind,
-    /// Boxed actual error for easy passing
-    pub error: Box<error::Error + Send + Sync>,
+    kind: ErrorKind,
+    error: Box<error::Error + Send + Sync>,
 }
 
 /// Types of errors produced in the linalg module.
 ///
 /// List intended to grow and so you should
 /// be wary of matching against explicitly.
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 pub enum ErrorKind {
     /// An argument did not uphold a necessary criteria for the function.
     InvalidArg,
--- a/src/error.rs
+++ b/src/error.rs
@@ -9,8 +9,10 @@ use std::marker::{Send, Sync};
 /// An error related to the linalg module.
 #[derive(Debug)]
 pub struct Error {
-    kind: ErrorKind,
-    error: Box<error::Error + Send + Sync>,
+    /// Type of error that led to this issue
+    pub kind: ErrorKind,
+    /// Boxed actual error for easy passing
+    pub error: Box<error::Error + Send + Sync>,
 }
 
 /// Types of errors produced in the linalg module.
@@ -25,6 +27,8 @@ pub enum ErrorKind {
     DecompFailure,
     /// A failure due to some algebraic constraints not being met.
     AlgebraFailure,
+    /// Tried to divide by zero
+    DivByZero
 }
 
 impl Error {
--- a/src/matrix/decomposition.rs
+++ b/src/matrix/decomposition.rs
@@ -1496,4 +1496,18 @@ mod tests {
 
         let _ = a.lup_decomp();
     }
+
+    #[test]
+    fn test_lup_decomp() {
+        use error::ErrorKind;
+        let a = Matrix::<f64>::new(4, 4, vec![
+                                   1., 2., 3., 4., 
+                                   0., 0., 0., 0., 
+                                   0., 0., 0., 0., 
+                                   0., 0., 0., 0.]);
+        match a.lup_decomp() {
+            Err(e) => assert!(*e.kind() == ErrorKind::DivByZero),
+            Ok(_) => panic!()
+        }
+    }
 }
--- a/src/matrix/decomposition.rs
+++ b/src/matrix/decomposition.rs
@@ -1157,8 +1157,8 @@ impl<T> Matrix<T> where T: Any + Copy + One + Zero + Neg<Output=T> +
                 let denom = u[[i,i]];
 
                 if denom == T::zero() {
-                    return Err(Error::new(ErrorKind::DecompFailure,
-                        "Matrix could not be LUP decomposed."));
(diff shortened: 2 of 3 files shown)


andrewcsmith:
Also, I made the error fields public because there doesn't seem to be a clear reason why they're not public, and also because they can now be destructured in a match statement.

src/matrix/mod.rs
```diff
@@ -541,7 +541,11 @@ impl<T: Any + Float> Matrix<T> {
             (self[[0, 1]] * self[[1, 0]] * self[[2, 2]]) -
             (self[[0, 2]] * self[[1, 1]] * self[[2, 0]])
         } else {
-            let (l, u, p) = self.lup_decomp().expect("Could not compute LUP decomposition.");
+            let (l, u, p) = match self.lup_decomp() {
+                Ok(x) => x,
+                Err(Error { kind: ErrorKind::DivByZero, .. }) => return T::zero(),
+                _ => { panic!("Could not compute LUP decomposition."); }
```
AtheMathmo: This is definitely a welcome change!

I wonder if it is still worth us panicking here (instead of returning a `Result`). I think LUP can only fail in this case and so we should never see the panic. If that assumption is correct then I think keeping the panic is appropriate.

src/matrix/decomposition.rs
```diff
@@ -1157,8 +1157,8 @@ impl<T> Matrix<T> where T: Any + Copy + One + Zero + Neg<Output=T> +
                 let denom = u[[i,i]];
 
                 if denom == T::zero() {
-                    return Err(Error::new(ErrorKind::DecompFailure,
-                        "Matrix could not be LUP decomposed."));
+                    return Err(Error::new(ErrorKind::DivByZero,
+                        "A value in the diagonal of U == 0.0."));
```
AtheMathmo: When this error is propagated will `U` make sense to the reader? Maybe it's better to say: `Singular matrix found in LUP decomposition`?

src/error.rs
```diff
@@ -9,8 +9,10 @@ use std::marker::{Send, Sync};
 /// An error related to the linalg module.
 #[derive(Debug)]
 pub struct Error {
-    kind: ErrorKind,
-    error: Box<error::Error + Send + Sync>,
+    /// Type of error that led to this issue
+    pub kind: ErrorKind,
```
AtheMathmo: I'm not sure if there _are_ any negative consequences to making these public - but really we only want the fields to be readonly. Is it not possible to achieve the same effect you wanted by adding a new function to retrieve the `error`? (One already exists for the `kind`).

Either way this change is welcome.

src/matrix/mod.rs
andrewcsmith: Sure, I'm just trying to match the original method as closely as possible. Returning a `Result` would require changing the function signature, which would definitely be a breaking change. I do think it would improve the interface though.

AtheMathmo: Ok, it seems it is worth having an issue + discussion over. Let's keep this as-is for the merge.

src/error.rs
andrewcsmith: Turns out, yes, it is possible. (Becoming a Pattern Match Wizard is on the todo list I guess.)

src/matrix/decomposition.rs
andrewcsmith: I wondered that myself. Can we include both? I like the specificity in that it gives people some idea of how to handle the error (which is recoverable, in the case of the determinant).

src/matrix/mod.rs
andrewcsmith: On second thought, perhaps "change all possible panics to results" should be a separate issue associated with one breaking change. There are *lots* of places in this code that I would rather see a `Result`, particularly because the `panic` crashes a thread and I can't have that in my particular use-case.

AtheMathmo: Yeah this is definitely still an issue - when I first started I was _only_ using panics but things have improved a lot.

I'd appreciate you opening up an issue and listing some of these that you think need improving.

src/matrix/decomposition.rs
AtheMathmo: I think including both looks good.

AtheMathmo:
I'm happy to merge this now if you are?

Andlon:
May I suggest to also add test(s) for this behavior, to prevent future regressions...?

AtheMathmo:
@Andlon right you are! Thank you.

andrewcsmith:
fair point. here ya go.

src/matrix/mod.rs
```diff
@@ -952,6 +952,14 @@ mod tests {
         println!("det is {0}", f);
         let error = abs(f - 99.);
         assert!(error < 1e-10);
+
+        let g = Matrix::<f64>::new(4, 4, vec![
+                                   1., 2., 3., 4., 
+                                   0., 0., 0., 0., 
+                                   0., 0., 0., 0., 
+                                   0., 0., 0., 0.]);
```
Andlon: You couldn't have known this, but I think it was [recently decided](https://github.com/AtheMathmo/rulinalg/pull/47#issuecomment-248436965) to start writing tests with the (new) `matrix!` macro, because it gives us some compile-time safety and is perhaps somewhat more intuitive.

Not to say that this needs to be done now, though. I'll leave it to @AtheMathmo if he wants to enforce this already (even though the issue that was mentioned in the previous link does not exist yet).

In any case, thanks for adding the tests! Makes me sleep better at night :P

AtheMathmo: If @andrewcsmith is happy to make this change I think it is good to do now.

I'll also write up a quick issue before I forget again - thanks!

andrewcsmith: got it - done

AtheMathmo:
This looks good to me - everyone else happy to merge? @Andlon

Andlon:
Aye, aye, captain!

Looks good to me :)

PR #55, opened 2016-09-27
Diagonal iterators

### LINKED ISSUE #53: Allow Matrix diagonal as mutable elements
We could reduce the number of places in the code with `get_unchecked` calls if we were able to return a `MatrixSlice` and a `MatrixSliceMut` that consists of a diagonal.

This would only require setting `row_stride` equal to `cols() + 1`, so that the column is incremented each time. We should have `assert!` checks to ensure that the inputs are good, then use `get_unchecked_mut` if all is well.

AtheMathmo:
This PR introduces new iterators (`Diagonal`, `DiagonalMut`) for `Matrix`/`MatrixSlice`/`MatrixSliceMut`. It is intended to resolve #53 .

Here are some benchmarks for collecting the main diagonal of a matrix:

```
test linalg::iter::mat_diag_iter_100_500           ... bench:         272 ns/iter (+/- 34)
test linalg::iter::mat_diag_manual_100_500         ... bench:         309 ns/iter (+/- 87)
test linalg::iter::mat_diag_iter_10_50             ... bench:          62 ns/iter (+/- 8)
test linalg::iter::mat_diag_manual_10_50           ... bench:          68 ns/iter (+/- 21)
```

The manual version uses unsafe `get_unchecked` indexing _and_ is slower!

Following is kept in place for prosperity but this PR is no longer WIP (pending review).

---

This PR is currently a work in progress and the following should be addressed:

- How do we handle iteration of off-diagonals?
- How do we expose these iterators?

I'm thinking we do the following:

Add the following:

```rust
pub enum DiagOffset {
    Above(usize),
    Below(usize)
}
```

The iterator will then have this enum as a field - which will determine which diagonal offset we use. When creating we will have to check that `Above(x)` has `x < mat.cols()`, and `Below(y)` has `y < mat.rows()`.

In terms of exposing the iterator - I think the following is a good approach:

- Add `fn iter_diag(&self, k: DiagOffset)` to `BaseMatrix`.
- Add `fn iter_diag_mut(&mut self, k: DiagOffset)` to `BaseMatrixMut`.

We will keep the `diag` function as returning a `Vec` for now - mostly to prevent breaking changes and we can reassess in future.

We will probably need to move the iterators into `matrix/mod` (they are currently in `matrix/iter`). This is so that we can construct them in `matrix/slice`.


src/matrix/iter.rs
```diff
@@ -210,14 +300,149 @@ impl<'a, T> IntoIterator for &'a mut MatrixSliceMut<'a, T> {
...
+    impl<'a, T> Diagonal<'a, T> {
+        fn from_mat<M: BaseMatrix<T>>(mat: M) -> Diagonal<'a, T> {
+            Diagonal {
+                slice_start: mat.as_ptr(),
+                diag_pos: 0,
+                diag_len: cmp::min(mat.rows(), mat.cols()),
+                row_stride: mat.row_stride() as isize,
+                _marker: PhantomData::<&'a T>,
+            }
+        }
```
Andlon: I presume you're doing this so you can test the feature without being intrusive into the `Matrix` types...?

AtheMathmo: Yes - I will almost definitely remove these helped functions once the iterators are properly exposed in the API.

Andlon:
Had a brief look. Looks good to me!

One comment though. Should we also have a `DiagonalOffset::Main` for the sake of symmetry? It feels very arbitrary to pick `Above(0)` or `Below(0)` if you want the main diagonal (which is probably what users want in the majority of cases).

Also, I generally like this pattern better than the signed integer approach, but I'm curious if there are algorithmic patterns which are not easily covered by the more type-safe approach you propose. I can't think of any, so perhaps we should just go ahead and re-assess if we receive any complaints.

If you recall [my previous example using signed integers](https://github.com/AtheMathmo/rulinalg/issues/53#issuecomment-249661330), I think actually it looks much more clear this way!

```
use DiagonalOffset as D;
for diag in (D::Above(1), D::Main, D::Below(1) {
    for ref mut element in mat.iter_diag_mut(diag) {
        *element *= scale_factor;
    }
}
```

AtheMathmo:
In the issue thread I had suggested this - and thought it looked a little ugly, but actually now I think it would be better. I'll use the enum with the `Main` entry too.

I also can't think of any cases where this approach fails and the signed one wins. As you say I guess we'll implement it and respond to complaints later :).

Andlon:
Also, I understand that you want to avoid a breaking change right now, but I _really_ think that `diag` should return an iterator by default in the future! For one, because I think in many (most?) cases you don't even need to collect into a separate collection, you just want to iterate over the values anyway. Secondly, because you can collect it into different data structures (`Vector`, `Vec`, etc.), or even `.chain()` or `.zip()` different diagonals together, or similar interesting patterns. Third, I guess we want to keep the API smaller, and as such we avoid somewhat redundant methods.

Andlon:
A question.

How should we respond if the user asks for a diagonal offset which is out of bounds? For example, `Above(2)` for a `3x3` matrix? I guess the most comparable in the standard library is `slice.get(idx)`, which returns an `Option<T>`. So we _could_ return an `Option<Diagonal>`, but this is very inconvenient when you are only interested in `Main`, in which case the diagonal *always* exists. One possible approach here is to remove `Main` from the enum again, and have separate methods, i.e.

```
/// Iterator over the diagonal
fn diag_iter(&self) -> Diagonal;
/// Iterator over off-diagonals
fn off_diag_iter(&self, offset: DiagonalOffset) -> Option<Diagonal>;
```
It feels a little verbose, but it's clean, at least. I'm not sure what the best approach here is.

AtheMathmo:
I was planning on returning an `Option<Diagonal>` even for the `Main` diagonal. I agree that it is a little unnerving for the `Main` case but _most_ of the time when we have to unwrap an `Option` we are _sure_ that the value does exist. I don't think this goes against that too strongly - I'm happy to discuss a little more and see what the finished product looks like though.

Andlon:
I looked a bit more at `std::slice`. I'd say panicing is acceptible for invariant-breaking input, see for example [slice.split_at()](https://doc.rust-lang.org/std/primitive.slice.html#method.split_at).

Andlon:
There is a conceptual difference in returning `Option<Diagonal>` for the main diagonal and `Diagonal` directly. In the first case, the user needs to actually sit down and think: "Is it safe to unwrap this?", whereas if we provide a `Diagonal` directly, the user knows that it is indeed safe to unwrap.

I think we should definitely strive to not impose redundant considerations for our users whenever possible!

AtheMathmo:
Actually - I think the most idiomatic approach would be to return an empty iterator. And I think we can achieve that by setting the `diag_len` to zero if it is out of bounds. I'll try implementing that so we can see how it looks.

Otherwise I'm happy to go with the panicking approach (which will be low effort to change).

Andlon:
I think this might be dangerous. I think in most cases if you try to get a diagonal that does not exist, this is a bug in your code. Depending on your subsequent actions, you can get very unexpected results that will happily go along without letting you know! For example, consider the following example:

We wish to compute the sum of the off-diagonal elements (the trace) of some off-diagonal. We do the following:

```
use DiagonalOffset::Above;
let sum = mat.iter_diag(Above(some_num)).sum();
```
This is very simple and convenient! But what if `some_num` is out of bounds? We only get a zero sum, with no warning whatsoever that something went wrong!

I'd rather go with the panic or the `diag/offdiag` split, I think.

AtheMathmo:
Also I think I agree with your points. Returning an empty iterator could be a little confusing and is very difficult to detect (without consuming the iterator).

I would prefer to go with panicking for now - the rulinalg api already feels quite large to me and so I'm always anxious to add lots of similar functions. (This is an issue we'll hopefully find a way to address in _the future_).

src/matrix/iter.rs
```diff
@@ -1,9 +1,94 @@
...
+    }
+
+    fn nth(&mut self, n: usize) -> Option<Self::Item> {
+        if self.diag_pos + n < self.diag_len {
+            let row = (self.diag_pos + n) as isize;
+            let offset = match self.diag_offset {
+                DiagOffset::Main => row * self.row_stride + row,
+                DiagOffset::Above(k) => row * self.row_stride + row + k as isize,
+                DiagOffset::Below(k) => row * self.row_stride + row - k as isize,
```
AtheMathmo: This is definitely wrong - and some of the other `Below` implementations may be as well.

Andlon:
Agree with your assessment. I think also panicking is perfectly acceptable, and also my preference.

AtheMathmo:
I still need to add tests for the slice `nth` and `last` functions. Then write up some docs explaining how these things work.

After that I'll remove the WIP tag!

src/matrix/iter.rs
```diff
@@ -210,14 +295,177 @@ impl<'a, T> IntoIterator for &'a mut MatrixSliceMut<'a, T> {
...
+
+        // Check equivalence
+        for ((d1, d2), d3) in a.iter_diag(DiagOffset::Main)
+            .zip(a.iter_diag(DiagOffset::Above(0)))
+            .zip(a.iter_diag(DiagOffset::Below(0))) {
+            assert_eq!(d1, d2);
+            assert_eq!(d1, d3);
+        }
+        let diags = [0.0, 4.0, 8.0];
+        assert!(a.iter_diag(DiagOffset::Main).enumerate().all(|(i, diag)| diags[i] == *diag));
```
Andlon: Can I maybe suggest instead collecting the `iter_diag` and using `assert_eq!` against `diags` instead? Would make the test a lot easier to read!

src/matrix/iter.rs
```diff
@@ -210,14 +295,177 @@ impl<'a, T> IntoIterator for &'a mut MatrixSliceMut<'a, T> {
...
+                            6.0, 7.0, 8.0];
+
+
+        // Check equivalence
+        for ((d1, d2), d3) in a.iter_diag(DiagOffset::Main)
+            .zip(a.iter_diag(DiagOffset::Above(0)))
+            .zip(a.iter_diag(DiagOffset::Below(0))) {
+            assert_eq!(d1, d2);
+            assert_eq!(d1, d3);
+        }
```
Andlon: Perhaps this should be a separate test? Also, perhaps write it out explicitly if it's not too much work, as it's not immediately obvious what's going on (I mean, it is after reading it, but if a (unit) test is not immediately obvious, it's all the more likely for bugs to creep into the test itself)?

src/matrix/iter.rs
AtheMathmo: Yes - and also ensure our iterators aren't terminating early... Good call!

src/matrix/iter.rs
AtheMathmo: Maybe this is worth having as a separate test. At first I was thinking of it as a kind of validation but it is actually testing its own case independently.

AtheMathmo:
I think this is ready for a proper review now. Also I've added benchmarks (on my system) to the PR description.

Andlon:
From a quick look, it looks good, and the benchmarks are awesome! I have a couple of things I'm curious about though. Will do a proper review tonight!

Andlon:
@AtheMathmo: one thing though, what _exactly_ is `row_stride`? The docs only say "Row stride in the matrix". I can guess, but it would be nice to have a precise definition.

AtheMathmo:
From my understanding: _row stride_ is a conventional name for the number of elements in a row in your underlying data structure. It's needed because if we have a matrix with 5 columns and 3 rows, but we take a slice with only 3 columns then we need to know that our underlying `Vec` actually has 5 elements in each matrix row, that is the row stride.

We should document this a little more. Though from memory the only place that it is exposed to the user right now is in the `from_raw_parts` function.

Andlon:
That makes sense. Thanks! Yeah, some documentation on that would be nice, even if it's just internal for developers. (That might exist already though, I didn't look through all the source)

tafia:
Manual is faster on my system. 

Any reason why you didn't borrow a BaseMatrix to store Matrix data (like row_stride and the pointer)?

I will try an alternative, can you wait for 1 or 2 days?

AtheMathmo:
> Manual is faster on my system.

Thanks for letting me know! Is it by a considerable amount? Maybe you could try some different sizes in case it is cache differences?

> Any reason why you didn't borrow a BaseMatrix to store Matrix data (like row_stride and the pointer)?

I just went with pointers directly as it was the approach I had taken before. In terms of safety they should be mostly equivalent as we have lifetimes etc. There is a little more room for error in that I am manually computing the offsets when we could be using `get_unchecked` - this is fairly minor though (I would argue). [As an aside we should probably look into inlining those `get` functions].

There may also be some borrowing issues when we use a `BaseMatrixMut` but I haven't tried it out so can't say for sure.

> I will try an alternative, can you wait for 1 or 2 days?

Sure - I'm happy to wait.

src/matrix/iter.rs
```diff
@@ -1,9 +1,99 @@
...
+            unsafe {
+                // Get pointer and shift to current diagonal
+                let offset = match self.diag_offset {
+                    DiagOffset::Main => self.diag_pos as isize * self.row_stride
+                                        + self.diag_pos as isize,
+                    DiagOffset::Above(k) => self.diag_pos as isize * self.row_stride
+                                            + (self.diag_pos + k) as isize,
+                    DiagOffset::Below(k) => (self.diag_pos + k) as isize * self.row_stride
+                                            + self.diag_pos as isize,
+                };
```
Andlon: Instead of dealing with the DiagOffset on every `next()` invocation, couldn't you deal with it once during construction? I'll just sketch out what I mean below.

```
pub struct Diagonal<'a, T: 'a> {
    matrix: &'a BaseMatrix<T>
    start: [usize; 2],
    diag_pos: usize,
    diag_len: usize
 }

impl<'a, T> Iterator for Diagonal<'a, T> {
    type Item = &'a T;
    fn next(&mut self) -> Option<Self::Item> {
        if self.diag_pos < self.diag_len {
            let index = [ self.start[0] + self.diag_pos, self.start[1] + self.diag_pos];
            let value = unsafe { self.matrix.get_unchecked(index) };
            self.diag_pos += 1;
        } else {
            None
        }
    }

    // And the rest similarly
}
```

Then the offset is the only difference between different values of `DiagOffset`, and the `Iterator` implementation doesn't care which diagonal it is iterating over.

tafia:
Something like https://github.com/AtheMathmo/rulinalg/pull/59

src/matrix/iter.rs
Andlon: Thinking about it, if `BaseMatrix` had a `get(&self) -> Option<T>` method, it might even be possible to implement it efficiently entirely in safe Rust:

```
pub struct Diagonal<'a, T: 'a> {
    matrix: &'a BaseMatrix<T>,
    start: [usize; 2],
    diag_pos: usize
 }

impl<'a, T> Iterator for Diagonal<'a, T> {
    type Item = &'a T;
    fn next(&mut self) -> Option<Self::Item> {
        let index = [ self.start[0] + self.diag_pos, self.start[1] + self.diag_pos];
        let result = self.matrix.get(index);
        self.diag_pos += 1;
        result
    }

    // And the rest similarly
}
```
Sure, `get` does bounds check, but you'd have to do that anyway with the `diag_pos`. I wouldn't be too surprised if this approach is just as fast as any pointer-offset approach.

AtheMathmo:
Thank you both for the feedback!

I agree that it makes a lot of sense to borrow the `BaseMatrix` and use this for the matrix fields. And also to modify the start point rather than having branches within each of the iterator functions.

I have some concerns about how well @tafia 's approach (and the `get` approach @Andlon suggested) will work with the compiler. In the current implementation the compiler seems to magically remove a lot of the bound checks from the `diag_pos < diag_len`. Hopefully this is still the case with the suggested versions but if not it isn't a big deal.

I'll implement these changes and rerun the benches to confirm!

Andlon:
I think in any case, unless really possible to avoid because of performance reasons, I think we should try to avoid working directly with pointer offsets, as it exposes/breaks the encapsulation of how matrices are stored internally (which is a maintenance burden).

AtheMathmo:
I agree - if the mutable iterator plays nice with borrowing the `BaseMatrixMut` I think your version is equivalent? And if @tafia 's version works too then it means our iterator allocates less (we only need one usize for the start and one for the end).

I'm trying your version first - if that works I'll try it with `get` instead of `get_unchecked`. And then if all is good @tafia's version in #59 should be optimal!

AtheMathmo:
@Andlon  We _have_ to use raw pointers sadly. From #59:

> rustc can't be sure that our iterator isn't returning two mutable references to the same point. Raw pointers it is!

I will merge in @tafia's PR once I can confirm there are no regressions on my machine.

Andlon:
@AtheMathmo: ah, yes. That makes sense! Fair enough :)

I found this: http://smallcultfollowing.com/babysteps/blog/2013/10/24/iterators-yielding-mutable-references/. It's from 2013, but it appears it is the same basic problem, and I don't expect there's any better solution available yet.

src/matrix/iter.rs
```diff
@@ -1,9 +1,76 @@
...
+/// Iterates over the rows in the matrix.
+impl<'a, T, M: $diag_base<T>> Iterator for $diag<'a, T, M> {
+    type Item = $diag_type;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.diag_pos < self.diag_end {
+            let pos = self.diag_pos as isize;
+            self.diag_pos += self.matrix.row_stride() + 1;
+            unsafe {
+                Some(self.matrix.$as_ptr().offset(pos).$to_item().unwrap())
```
AtheMathmo: I _think_ this is the right approach but I thought I'd raise the question.

Shall we just return the `Option` from `$to_item`? I'd argue **no** as we'd prefer to panic if we accidentally stumble upon a null pointer.

Andlon: Agree with your assessment. If there is a null pointer, it's a bug in our implementation. Perhaps replace `unwrap` with an `expect` with a short message explaining this...?

AtheMathmo:
I am happy to merge this now - am I missing anything else @tafia @Andlon ?

src/matrix/iter.rs
```diff
@@ -1,9 +1,85 @@
...
+            None
+        }
+    }
+
+    fn last(self) -> Option<Self::Item> {
+        if self.diag_pos < self.diag_end {
+            unsafe {
+                Some(self.matrix.$as_ptr()
+                                .offset(self.diag_end as isize - 1)
+                                .$to_item()
```
Andlon: I'm not at all familiar with raw pointers, but if we're already so deep into unsafe territory, couldn't we just dereference the pointer at this point, rather than calling `.as_ref()` or `.as_mut()`, since this (I think?) might incur a possible null check? I don't know if the compiler will elide this check or not. If it has no impact on performance it's better the way it is though!

src/matrix/mod.rs
```diff
@@ -95,6 +95,35 @@ pub struct RowsMut<'a, T: 'a> {
...
+
+/// An iterator over the diagonal elements of a matrix.
+#[derive(Debug)]
+pub struct Diagonal<'a, T: 'a, M: 'a + BaseMatrix<T>> {
+    matrix: &'a M,
+    diag_pos: usize,
+    diag_end: usize,
+    _marker: PhantomData<&'a T>,
+}
+
```
Andlon: Do you really need separate types `M` and `T`, or could you get by with only `T` and having a reference to a `BaseMatrix<T>`, in which case you could also safely drop the `PhantomData`...?

I'm not saying you have to change this, but I'm actually quite curious!

Andlon:
I think it looks good! I added a couple of comments mostly out of curiosity. They don't need to impede the merge.

src/matrix/mod.rs
AtheMathmo: If we wanted to use it as a trait object I think we would need to `Box` it? This means an extra heap allocation which in practice is probably negligible but I think I'd rather add the extra generic type/marker.

Andlon: Ah, yeah, that might be! I've somehow managed to not touch upon trait objects too much so far, so I don't have much experience in that field. Sounds good :)

src/matrix/iter.rs
AtheMathmo: So we can't do a proper dereferencing in the macro. This is because we can't differentiate between `&*(raw_ptr)` and `&mut *(raw_ptr)` (if someone knows how I'd be happy to hear it).

What we _can_ do however is use `mem::transmute`. It wouldn't be the first time I've used it in rulinalg but it's a pretty scary thing that we should probably try to avoid... I did test it out with `mem::transmute` and saw the exact same performance - I guess this is good enough motivation to keep the unwrapped _redereferencing_!

Andlon: Haha, all right, all right! Then all is good :)

src/matrix/mod.rs
AtheMathmo: > I've somehow managed to not touch upon trait objects too much so far

Neither have I - they seem pretty tricky and so I prefer to avoid them...

src/matrix/iter.rs
AtheMathmo: Actually... maybe it is worth using `mem::transmute` _just in case_ it gives a performance boost on other systems. I'd be interested to know if this is the case for you or @tafia - if either of you have the time could you post up the benchmarks with the `next` function modified to use mem::transmute?

bright-star:
Wow, perfect timing. I'm itching to use this.

Two dissenting opinions:

- Is it culturally such an issue to return an `Option` or `Result` instead of panicking for a Rust programmer? Theoretically speaking, a matrix with zero rows and columns (not a mathematically implausible concept) would have no main diagonal, so it's not unreasonable to make that case realizable using Rust patterns. I think that not doing this externalizes the effort of wrapping the result of the function call--every programmer who has to plan for unpredictable input to this function has to do the same effort.
- Regarding the signed super/subdiagonal differentiation, I agree it's more idiomatic to Rust to use an `enum` than overloading the signed integers to describe this. (Although, technically, *all* programming is at heart an overloading of the integers when you think about it.) It's worth mentioning that any user who wants to do interesting numeric patterns (say, quick alternation across sub/superdiagonals, or something even stranger) in populating a matrix will indeed have to spend extra effort to map some pair of numbers to `Above` and `Below`.

AtheMathmo:
Thanks for the input @bright-star !

With regards to your points:

1. I am also a little unsure on this one - but I'm leaning towards panicking being the right course of action. Generally a panic represents a programmer error which I think in this case makes sense: the programmer has gone out of bounds and we should let them know. We should check this is the case - but I believe currently an empty (0 rows/cols) matrix will return an empty iterator if you request `Main`, and panic otherwise. This might be a little inconsistent but is reasonable I think.
2. I think as you discuss this is really a trade-off between simplicity for 99% of our users vs. added complication for the remaining 1%. I think in this case the added complication is reasonable and Rust makes it fairly easy to do this enum magic (`match`, `map` etc.). It is still a worthwhile consideration though.

Andlon:
@bright-star: good points! The zero-diagonal is an edge case we should probably have a test for, and also perhaps explicitly describe the behavior in the documentation, although it already says it will never panic in the case of `DiagOffset::Main`. This also sort of makes sense if you define a 0x0 matrix to have a 0-length main diagonal. I think iterating over the other diagonals is relatively rare in comparison to the main diagonal, and even rarer in the cases where you have to deal with unpredictable input. Like @AtheMathmo mentioned for point 2, I guess it's a tradeoff in simplicity for the many vs flexibility for the few. That said, we might be making the wrong decision here, but perhaps we could just see what kind of feedback we get from users when it's more used in actual applications?

As for point 2, we _could_ simply `impl From<integer> for DiagOffset` I guess, and take `Into<DiagOffset>` instead of `DiagOffset` in `diag_iter`? It might actually be such a common case that it's still worth supporting it. I don't know.

AtheMathmo:
@Andlon I actually really like the idea of `Impl From<integer> for DiagOffset`. I'll want to see what it looks like once it's there but it seems like it would make things a little simpler if the user doesn't _have_ to know about the enum.

bright-star:
Just for reference, here is some code I'm porting over from MATLAB that uses diagonal addressing: https://github.com/bright-star/computational-em/tree/master/finite-difference

AtheMathmo:
@bright-star thanks to your post we caught a bug with empty matrices, thank you!

What did you think of the idea of implementing `From<integer>` for `DiagOffset`? Does this alleviate your concerns about playing with more interesting numeric patterns?

bright-star:
Thanks! I'm fine with that. It's definitely something that should be the responsibility of the library user to decide if they want to use.

bright-star:
Is there anything I can do to help this PR get merged?

AtheMathmo:
@bright-star Sorry I've been a little slow with this as I am currently on vacation. I return on Tuesday so should be able to settle it shortly after that.

As for the delays - I still need to implement `From<integer>` for `DiagOffset`. Otherwise there is another PR, #61 which proposes a tidier implementation. However this seems to have some performance regression and so I would prefer to leave the PR as is. @tafia are you ok for me to close that? I've played with it a little - switching to `mem::transmute` etc. but still see the regression.

I also switched this PR to use `mem::transmute`. Though I saw no performance difference it removes the null check and so seems worth including...

tafia:
@AtheMathmo, yes you can close #61. Better to have the iterator available sooner than later as changing how it is implemented, in case we managed to understand the regression, won't be a breaking change.

AtheMathmo:
@tafia Thanks, I agree though it is a shame. I think your current contributions to this PR have already made a lot of difference though!

I've implemented the conversion to DiagOffset so that users can now directly use integers on the `iter_diag` functions. I think this is ready to merge now but please let me know if anyone sees anything off.

AtheMathmo:
@Andlon - do you mind if I poke you for confirmation on merging this?

I'd like to get it in soon. It seems to be a popular feature!

src/matrix/slice.rs
```diff
@@ -1040,7 +1041,7 @@ pub trait BaseMatrixMut<T>: BaseMatrix<T> {
     ///                 3, 4, 5;
     ///                 6, 7, 8];
     /// // Zero the sub-diagonal (sets 3 and 7 to 0)
-    /// for sub_d in a.iter_diag_mut(DiagOffset::Below(1)) {
+    /// for sub_d in a.iter_diag_mut(-1) {
```
Andlon: Maybe not a very big deal, but I'm of the opinion that the preferred way of using `iter_diag[_mut]` should be through using the `DiagOffset` enum, as it is type-safe and less prone to error (negative sign can very well be interpreted as both above or below the diagonal, it's really down to convention), with the signed integer argument being provided as a way for people to implement certain specific patterns that are easily written using signed integers but would otherwise be clunky to implement with the enum.

What's your take on this?

src/convert.rs
```diff
@@ -40,9 +42,41 @@ impl_matrix_from!(MatrixSlice);
 impl_matrix_from!(MatrixSliceMut);
 
 
+macro_rules! impl_diag_offset_from {
+    ($unsigned_type:ty, $signed_type:ty) => {
+impl From<$unsigned_type> for DiagOffset {
```
Andlon: Do we actually want to implement `From<unsigned>`? It doesn't personally make a lot of sense to me to ever use an unsigned integer for this (i.e. then you should *really* rather use `DiagOffset::Above`).

src/matrix/slice.rs
bright-star: `DiagOffset` enum matches above discussion.

Andlon: It does, but I'm referring to the documentation here (sorry, may not have been clear).

Andlon: Yeah, I think perhaps having both in the docs would be good. Show main usage with `DiagOffset`, and then also mention that it is possible to use integers instead (but somehow make it clear that this is the exception, not the rule).

src/convert.rs
AtheMathmo: I agree partly, but I'm not too sure if this is a nice experience for the user when removed. It seems unwieldy to allow the user to do: `mat.iter_diag(-1)` but then need `mat.iter_diag(1i32)` is they prefer this pattern.

The current conversion maps any unsigned int to an `Above` offset (or `Main` if 0). I don't see clearly the disadvantage here? If the argument is that we'd prefer our users to understand what they're doing and use the enum to show that absolutely, then I do see the merit but I think the solution should be removing the conversions entirely instead of just those for unsigned ints.

I'd really like to hear your thoughts on it though - it seems I may be missing something.

AtheMathmo:
Thanks for the comments - I gave responses. I think it's better to converge somewhat on these thoughts before we merge :)

src/convert.rs
Andlon: Ah, right. I thought Rust would infer the type of `1` to be a signed integer when used in this context. In that case I totally agree that it doesn't make sense to force the user to have to write `1i32`.

I was also thinking whether it would be better to remove the `Into`. In this case, users would have to write `.iter_diag(DiagOffset::from(k))` for a signed integer `k`. It's slightly more verbose, but it would encourage users to use the enum, while still being compatible with numeric patterns for the diagonals. I think `iter_diag(Below(1))` is a lot clearer and less prone to error than `iter_diag(-1)`. Ultimately, I don't have a very strong opinion on this, and either solution (`Into` or no `Into`) is fine with me.

bright-star:
Something just occurred to me. Does `DiagOffset::Above(0)` address the main diagonal? Should it? I think it might be a safe thing to do, and likewise with `DiagOffset::Below(0)`.

AtheMathmo:
@bright-star yes that is how it works.

Sorry I haven't made much progress on this PR. I'm having a hard time deciding exactly how the API should look. I'm not too sure about removing the `From` implementation for unsigned ints. At the end of the day it doesn't matter much at all so I'll settle with the common ground.

bright-star:
Maybe https://github.com/AtheMathmo/rulinalg/pull/69 will help crystallize your thoughts :wink:

AtheMathmo:
I want to try and get this in today, it seems we have two points to discuss.

1. Should the `iter_diag` functions take an `Into<DiagOffset>`, or just a `DiagOffset`.
2. Should we implement `From<unsigned>` for `DiagOffset`.

For the first point, I have no issue personally with allowing the generic `Into` parameter. But similarly I don't think it is too aesthetically damning to remove it. I'm happy to get rid of it and have a more explicit parameter here.

For the second point, it is also fairly minor. Again personally I don't see any issue with allowing unsigned types to cast into `DiagOffset`s. I also see no huge issue with not doing so. The only thing holding me back on this is the worry that it might impact us later. We index across the board using `usize` right now - I worry that there may be some cases were a user wants to use a `usize` index with the `DiagOffset. I don't see an obvious use case, and there wouldn't be much of an issue with having the user cast into a signed integer anyway.

If there are no additional thoughts on this I'll act on both points, removing the `Into` parameter and revoking the `From` implementation for unsigned ints.

Andlon:
Will comment on this when I'm home in a few hours!

On 30 Oct 2016 8:56 a.m., "James Lucas" <notifications@github.com> wrote:

>

> I want to try and get this in today, it seems we have two points to
discuss.
>
> Should the iter_diag functions take an Into<DiagOffset>, or just a
DiagOffset.
> Should we implement From<unsigned> for DiagOffset.
>
> For the first point, I have no issue personally with allowing the generic
Into parameter. But similarly I don't think it is too aesthetically damning
to remove it. I'm happy to get rid of it and have a more explicit parameter
here.
>
> For the second point, it is also fairly minor. Again personally I don't
see any issue with allowing unsigned types to cast into DiagOffsets. I also
see no huge issue with not doing so. The only thing holding me back on this
is the worry that it might impact us later. We index across the board using
usize right now - I worry that there may be some cases were a user wants to
use a usize index with the `DiagOffset. I don't see an obvious use case,
and there wouldn't be much of an issue with having the user cast into a
signed integer anyway.
>
> If there are no additional thoughts on this I'll act on both points,
removing the Into parameter and revoking the From implementation for
unsigned ints.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
<https://github.com/AtheMathmo/rulinalg/pull/55#issuecomment-257137473>, or
mute the thread

Andlon:
1. I think this is not an issue of aesthetics, but encouraging our users to leverage type safety for more readable and less ambiguity. I think removing `Into` is the safer choice.
2. Well, since DiagOffset maps one-to-one to signed integers, I think it makes sense to be conservative and avoid the usize implementation for now, and then see if we get any requests from users due to the lack of `From<unsigned>`? Then we could imagine legitimate usecases. This means less risk, but I think the risk is very low in the first place, so I'm not too fussed about it either way.

In any case, I'm still OK with whatever decision you want to make here. I also think it's OK to merge at this point.

That said... Please forgive me for bringing this up, as it may further confuse us all and cause further delays, but I've been thinking about diagonal iterators from the perspective of user experience. You don't need to take the following into account for this PR, but I think it brings up some important points.

Imagine you are a user of `rulinalg`. I think it's not unreasonable to assume that 90+% of the use cases for diagonal iterators will be iterating the main diagonal. You first write

```rust
for el in mat.d<autocomplete>
```
Unless your IDE is very smart, `iter_diag()` will not show up in the autocomplete prompt. After looking at the docs, you find `iter_diag()`, and without looking thoroughly at the documentation, you write

```rust
for el in mat.iter_diag() {}
```

Obviously, this doesn't work, because you need a diagonal offset, so you try again:

```rust
// I personally don't even know in what module DiagOffset is in off the top of my head!
for el in mat.iter_diag(rulinalg::DiagOffset::Main) {}
```

If you get the module right, then you should be good to go. Now, I argue that this user flow is not optimal. A good user experience implies that the user is intuitively guided so that his first choice is the right choice whenever possible. Constrast the above with the following:

```rust
// Main diagonal, `diag_iter()` shows up immediately when autocompleting
for el in mat.diag_iter() {}

// Off-diagonals cover the remaining use cases
for el in mat.offdiag_iter(rulinalg::DiagOffset::Above(2)) {}
```

My argument is that the current proposal has two flaws:

1. The naming does not put the main feature in focus and does not play nice with autocomplete functionality.
2. Working with the main use case, which is iterating over the main diagonal, is overly cumbersome.

I know the two-function proposal has more or less been abandoned already, but the more I think about it, the more I think it makes for a better fit. Note that this is _solely_ because Rust does not allow default arguments, in which case a single function would suffice.

Again, I'm very sorry for dragging this up again! Feel free to ignore it.

AtheMathmo:
> Again, I'm very sorry for dragging this up again! Feel free to ignore it.

Not at all! I think it is well worth discussing.

I'll go with your approach for points 1 and 2. It seems the more popular choice :).

As for the `diag_iter` vs `iter_diag`. I don't disagree that the former may be a little clearer _and_ work better with IDEs. However making this change would probably also mean making changes elsewhere. I don't think it would make sense to have `diag_iter` and `iter_rows` for example.

I think you're right about the primary use case however my plan for tackling that is to modify the current `diag` function. I think it will be better to have `diag` return an iterator over the main diagonal. From here it is a small step for the user to retain the existing functionality: `mat.diag().collect::<Vec<_>>()`. Of course this is a breaking change and so should come with a version bump.

Andlon:
Ah, I didn't know about your plans for `diag`. Then I whole-heartedly agree!

As for `iter_rows`. Yes, I agree, and I did actually think about this too. In fact, I think all `iter_*` functions should be changed to `*_iter` (well, `iter_rows()`, would probably be `row_iter()`, but you get my point). I think this feature is ready to merge, but I strongly suggest we consider renaming the iterator functions in a later release (this would of course be an annoying breaking change).

AtheMathmo:
I have created #75 and tagged as part of the 0.4 release. I'll write up a similar issue for the `diag()` change too.

I've just pushed changes to tighten up some of the conversion generics. I made some minor doc tweaks to reflect that too. I think this is ready to merge now, let me know if you spot anything!

Andlon:
@AtheMathmo: great! I looked over the latest commit and it looks good to me :)

AtheMathmo:
Ok, lets get this in!

PR #56, opened 2016-09-28
Moved bounds out of generic arguments

nwtnian:
Motivation behind this is improved readability in source code as well as rust-docs.

It’s much quicker to spot implemented traits (`Mul<Vector<T>>`) and/or the implementing type (`Matrix<T>`) in declarations like the following:

```rust
impl<T> Mul<Vector<T>> for Matrix<T> where T: Copy + Zero + Mul<T, Output=T> + Add<T, Output=T>
```

than is it to spot them in

```rust
impl<T: Copy + Zero + Mul<T, Output=T> + Add<T, Output=T>> Mul<Vector<T>> for Matrix<T>
```

After all we’re usually reading Rust code from left to right.

Also consistency.

### DIFF
--- a/src/matrix/decomposition.rs
+++ b/src/matrix/decomposition.rs
@@ -25,7 +25,9 @@ use libnum::{One, Zero, Float, Signed};
 use libnum::{cast, abs};
 use epsilon::MachineEpsilon;
 
-impl<T: Any + Float> Matrix<T> {
+impl<T> Matrix<T>
+    where T: Any + Float
+{
     /// Cholesky decomposition
     ///
     /// Returns the cholesky decomposition of a positive definite matrix.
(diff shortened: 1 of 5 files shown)


src/matrix/decomposition.rs
```diff
@@ -312,8 +314,12 @@ impl<T: Any + Float> Matrix<T> {
...
 /// containing the singular values, `u` is the matrix of left singular vectors
 /// and v is the matrix of right singular vectors.
-fn correct_svd_signs<T>(mut b: Matrix<T>, mut u: Matrix<T>, mut v: Matrix<T>)
-    -> (Matrix<T>, Matrix<T>, Matrix<T>) where T: Any + Float + Signed {
+fn correct_svd_signs<T>(mut b: Matrix<T>,
+                        mut u: Matrix<T>,
+                        mut v: Matrix<T>)
+                        -> (Matrix<T>, Matrix<T>, Matrix<T>)
+    where T: Any + Float + Signed
+{
```
Andlon: Agree that this is much more readable, and I personally definitely prefer this.

However, in light of your other PR in #57, is having `{` on the next line also compatible with `rustfmt`? I was under the impression that Rust styles normally dictate having `{` on the same line?

AtheMathmo:
I personally prefer having everything inline if the signature is small enough - but the difficulty there is judging when the signature is small _enough_. This is a good change and we'll try to keep it this way for consistency now.

Thank you!

PR #57, opened 2016-09-28
Cleaned up formatting.

nwtnian:
Cleaned up formatting by running `$ rustfmt` on rulinalg.

### DIFF
--- a/src/matrix/decomposition.rs
+++ b/src/matrix/decomposition.rs
@@ -321,12 +321,11 @@ fn correct_svd_signs<T>(mut b: Matrix<T>, mut u: Matrix<T>, mut v: Matrix<T>)
     // -1, which might be significantly faster in corner cases if we pick the matrix
     // with the least amount of rows.
     {
-        let ref mut shortest_matrix = if u.rows() <= v.rows() { &mut u }
-                                      else { &mut v };
+        let ref mut shortest_matrix = if u.rows() <= v.rows() { &mut u } else { &mut v };
         let column_length = shortest_matrix.rows();
         let num_singular_values = cmp::min(b.rows(), b.cols());
 
-        for i in 0 .. num_singular_values {
+        for i in 0..num_singular_values {
             if b[[i, i]] < T::zero() {
                 // Swap sign of singular value and column in u
                 b[[i, i]] = b[[i, i]].abs();
@@ -347,17 +346,19 @@ fn sort_svd<T>(mut b: Matrix<T>, mut u: Matrix<T>, mut v: Matrix<T>)
 
     // This unfortunately incurs two allocations since we have no (simple)
     // way to iterate over a matrix diagonal, only to copy it into a new Vector
-    let mut indexed_sorted_values: Vec<_> = b.diag().into_vec()
+    let mut indexed_sorted_values: Vec<_> = b.diag()
+        .into_vec()
         .into_iter()
         .enumerate()
         .collect();
 
     // Sorting a vector of indices simultaneously with the singular values
     // gives us a mapping between old and new (final) column indices.
-    indexed_sorted_values.sort_by(|&(_, ref x), &(_, ref y)|
-        x.partial_cmp(y).expect("All singular values should be finite, and thus sortable.")
-         .reverse()
-    );
+    indexed_sorted_values.sort_by(|&(_, ref x), &(_, ref y)| {
+        x.partial_cmp(y)
+            .expect("All singular values should be finite, and thus sortable.")
+            .reverse()
+    });
 
     // Set the diagonal elements of the singular value matrix
     for (i, &(_, value)) in indexed_sorted_values.iter().enumerate() {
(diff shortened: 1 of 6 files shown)


src/matrix/decomposition.rs
```diff
@@ -1264,15 +1255,15 @@ mod tests {
             .map(|((u_col, singular_value), v_col)| (Vector::new(u_col), singular_value, Vector::new(v_col)));
 
         assert!(singular_triplets.by_ref()
-            // For a matrix M, each singular value σ and left and right singular vectors u and v respectively
-            // satisfy M v = σ u, so we take the difference
+        // For a matrix M, each singular value σ and left and right singular vectors u and v respectively
+        // satisfy M v = σ u, so we take the difference
```
Andlon: Seems like a case unhandled by `rustfmt`? In-line comments should be indented when chaining methods, imo! Not so important, but worth taking note and potentially make an issue in the `rustfmt` repo? What do you think?

Andlon:
I think this is definitely a good idea. @AtheMathmo: could we consider making `rustfmt` a standard part of the contribution process, perhaps...?

src/matrix/mod.rs
```diff
@@ -717,9 +717,7 @@ fn back_substitution<T, M>(m: &M, y: Vector<T>) -> Result<Vector<T>, Error>
             }
 
             let diag = *m.get_unchecked([i, i]);
-            if diag.abs() < T::min_positive_value() +
-                T::min_positive_value()
-            {
+            if diag.abs() < T::min_positive_value() + T::min_positive_value() {
```
AtheMathmo: Looks like this is for back substitution solving - to check if the matrix is singular. It probably exists in forward substitution too... We can use the handy new `MachineEpsilon` trait it seems :).

Andlon: So `2 * eps` should be a sufficiently accurate bound? I really have to get around to immersing myself more into the corner cases of floating point arithmetic soon...

AtheMathmo:
Thank you for doing this! I'm certainly a bit lax with rustfmt...

In terms of enforcing as part of the contribution process I think it might be a good idea but I'm not entirely sure how. Another issue is that contributors frequently have different settings and so our rustfmts conflict with eachother - we can solve this by having a config file in the project root though.

There are a couple issues I've had with rustfmt that we should probably bring to the attention of the developers...

- [Some comments get shifted to left of file](https://github.com/AtheMathmo/rulinalg/blob/master/src/matrix/iter.rs#L15)
- Some long lines cause the parser to break (this may be resolved now).

I'll merge this after your other PR.

src/matrix/mod.rs
AtheMathmo: I am assuming it would be yes - though really I have very little idea myself! There may be a more technically complete way to do this.

Andlon:
If we have such a config file, would it be very hard to get the CI to run `rustfmt` and check for discrepancies?

Andlon:
Ah, see [here](http://johannh.me/blog/rustfmt-ci.html). Seems very plausible, actually!

src/matrix/mod.rs
Andlon: I'm not sure, but there may be more accurate bounds when comparing with identically zero. Just want to make sure that we're not discarding matrices that are actually nonsingular as singular because our bound is too low!

Should do some research on this...

AtheMathmo:
We should definitely look into this - but I'm not sure if it is viable right now. I checked and for me the parser still breaks on long lines. Some of these lines we can't reduce as they are long urls to reference material.

Maybe we can get around this with a custom config..

src/matrix/mod.rs
AtheMathmo: I opened #58 - maybe my terminology is backwards in the title though...

Andlon:
> We should definitely look into this - but I'm not sure if it is viable right now. I checked and for me the parser still breaks on long lines. Some of these lines we can't reduce as they are long urls to reference material.
> 
> Maybe we can get around this with a custom config..

Yeah, I guess this is not the most pressing issue anyway. Perhaps we could establish a long-term tracking issue for something we'd like to converge towards?

AtheMathmo:
Yes we probably should. I'll add after we merge this (in case any more information comes from this).

@nwtnian I am happy to merge this PR as is - however on my local machine I'm seeing a lot more changes when I run rustfmt. Even if I add the option to wrap long comments my rustfmt still fails on some longer code lines and I see changes in many more files than those affected here.

Could you give a little more detail on your set up? Which version of rustfmt are you using and do you have a custom config?

AtheMathmo:
@nwtnian pinging you on this again.

There are some conflicts now from where this has been sitting. Once we can converge on the rustfmt settings used I'll be happy to merge this. The issue currently is that my rustfmt still modifies these files that you have changed.

AtheMathmo:
Am closing this as it seems abandoned to me - will happily reopen if requested. Am also creating a new issue to standardize our formatting.

PR #59, opened 2016-09-29
poc simpler diag iterator

tafia:
Alternative way of writing the diagonal iterator which is basically a step_by iterator.
This only covers the non mutable part as a proof of concept.

### DIFF
--- a/src/matrix/slice.rs
+++ b/src/matrix/slice.rs
@@ -20,8 +20,7 @@
 //! ```
 
 use matrix::{Matrix, MatrixSlice, MatrixSliceMut, Rows, RowsMut, Axes};
-use matrix::{DiagOffset, Diagonal};
-// use matrix::{DiagOffset, Diagonal, DiagonalMut};
+use matrix::{DiagOffset, Diagonal, DiagonalMut};
 use matrix::{back_substitution, forward_substitution};
 use vector::Vector;
 use utils;
@@ -208,7 +207,7 @@ pub trait BaseMatrix<T>: Sized {
             matrix: self,
             diag_pos: diag_start,
             diag_end: diag_start + (diag_len - 1) * self.row_stride() + diag_len,
-            _marker: PhantomData::<&T>,
+            _marker: PhantomData::<T>,
         }
     }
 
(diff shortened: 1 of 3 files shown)


AtheMathmo:
I see - I think that this is simpler! I'd say it would be better still to keep the diag_pos as an index and instead use the `get_unchecked`? This also makes it a lot easier to handle the logic for the `Above` and `Below` cases.

I'd also like to check that the `DiagonalMut` version still works. I think that it should but things can often be surprising with the slicing hack stuff we have now.

AtheMathmo:
Oh - and also I'm not sure how the compiler would handle the `diag_pos` jumps like this. It seems (at least on my machine) that the compiler can remove many of the bounds checks as `diag_pos` increases by one and the bound on `diag_len` is large enough. We'd have to see how the benchmarks are effected with this change. Maybe you have some numbers already?

tafia:
I got about the same perf than previous version.
I've added back the mutable iterator.

The best would clearly be to use the rust step_by iterator if it ever gets stabilised.

AtheMathmo:
That's great!

I think the only other concern is what @Andlon said in #55 :

> I think in any case, unless really possible to avoid because of performance reasons, I think we should try to avoid working directly with pointer offsets

I was trying to use the `get_unchecked` and `get_unchecked_mut` functions but I get lifetimes issues in the `DiagonalMut` case - I really have no idea why but I'm trying to figure it out. Would you mind trying to use these in your implementation? For now I guess you can just convert your diag_pos into a row-col indices and use these functions. If it compiles then we should check that this approach doesn't incur a performance hit and I'll merge this in.

src/matrix/slice.rs
```diff
@@ -1057,26 +1055,26 @@ pub trait BaseMatrixMut<T>: BaseMatrix<T> {
...
-            slice_start: self.as_mut_ptr(),
-            diag_pos: 0,
-            diag_len: diag_len,
-            row_stride: self.row_stride() as isize,
-            diag_offset: k,
-            _marker: PhantomData::<&mut T>,
+            matrix: self,
+            diag_pos: diag_start,
+            diag_end: diag_end,
+            _marker: PhantomData::<T>,
```
AtheMathmo: Also we _should_ use this marker to enforce the lifetime too: `PhantomData::<&mut T>` (and in the struct definition). It isn't really needed as the `matrix` field does this already - but [is best practice](https://doc.rust-lang.org/std/marker/struct.PhantomData.html#unused-lifetime-parameter).

tafia:
I'll unfortunately not be able to do it today. I agree that using pointers
is not nice and we should avoid it.

Eventually we could probably even remove the structs entirely and use
something like
```
let mat = self.sub_slice(...); // Get the relevant squared matrix
0..diag_len.map(|i| unsafe { mat.get_unchecked([i, i]) });
```

I'll try tomorrow if you haven't yet.

AtheMathmo:
No problem - thanks!

When `impl Trait` return types land we'll be able to clean up quite a lot of the code. But for now I don't think the structs are _too_ messy :)

AtheMathmo:
Looks like we're not allowed to use `get_unchecked_mut` as rustc can't be sure that our iterator isn't returning two mutable references to the same point. Raw pointers it is!

PR #61, opened 2016-09-30
Diag sub matrix

tafia:
Here is a much simpler but surprisingly slower version of the iterator (at least on my system).
I suspect it has been slowed down by MatrxSlice initialisation, but I can't see why exactly, perhaps we're doing something wrong elsewhere.
This is not intended to be merged but I wanted to show a simpler alternative **without using pointers**.

### DIFF
--- a/src/matrix/slice.rs
+++ b/src/matrix/slice.rs
@@ -206,9 +206,8 @@ pub trait BaseMatrix<T>: Sized {
         };
 
         Diagonal {
-            matrix: self.sub_slice(diag_start, diag_len, diag_len),
+            square_matrix: self.sub_slice(diag_start, diag_len, diag_len),
             inner: 0..diag_len,
-            _marker: PhantomData::<&T>,
         }
     }
 
@@ -1072,9 +1071,8 @@ pub trait BaseMatrixMut<T>: BaseMatrix<T> {
         };
 
         DiagonalMut {
-            matrix: self.sub_slice_mut(diag_start, diag_len, diag_len),
+            square_matrix: self.sub_slice_mut(diag_start, diag_len, diag_len),
             inner: 0..diag_len,
-            _marker: PhantomData::<&mut T>,
         }
     }
 
--- a/src/matrix/slice.rs
+++ b/src/matrix/slice.rs
@@ -66,7 +66,9 @@ pub trait BaseMatrix<T>: Sized {
     }
 
     /// Get a reference to a point in the matrix without bounds checking.
-    unsafe fn get_unchecked(&self, index: [usize; 2]) -> &T {
+    unsafe fn get_unchecked<'a>(&self, index: [usize; 2]) -> &'a T 
+        where T: 'a
+    {
         &*(self.as_ptr().offset((index[0] * self.row_stride() + index[1]) as isize))
     }
 
@@ -190,22 +192,21 @@ pub trait BaseMatrix<T>: Sized {
     ///
     /// This function will never panic if the `Main` diagonal
     /// offset is used. 
-    fn iter_diag(&self, k: DiagOffset) -> Diagonal<T, Self> {
+    fn iter_diag(&self, k: DiagOffset) -> Diagonal<T> {
         let (diag_len, diag_start) = match k {
-            DiagOffset::Main => (min(self.rows(), self.cols()), 0),
+            DiagOffset::Main => (min(self.rows(), self.cols()), [0, 0]),
             DiagOffset::Above(m) => {
                 assert!(m < self.cols(), "Offset diagonal is not within matrix dimensions.");
-                (min(self.rows(), self.cols() - m), m)
+                (min(self.rows(), self.cols() - m), [0, m])
             },
             DiagOffset::Below(m) => {
                 assert!(m < self.rows(), "Offset diagonal is not within matrix dimensions.");
-                (min(self.rows() - m, self.cols()), m * self.row_stride())
+                (min(self.rows() - m, self.cols()), [m, 0])
             },
         };
 
         Diagonal {
-            matrix: self,
-            start: diag_start,
+            matrix: self.sub_slice(diag_start, diag_len, diag_len),
             inner: 0..diag_len,
             _marker: PhantomData::<&T>,
         }
@@ -850,7 +851,9 @@ pub trait BaseMatrixMut<T>: BaseMatrix<T> {
     }
 
     /// Get a mutable reference to a point in the matrix without bounds checks.
-    unsafe fn get_unchecked_mut(&mut self, index: [usize; 2]) -> &mut T {
+    unsafe fn get_unchecked_mut<'a>(&mut self, index: [usize; 2]) -> &'a mut T
+        where T: 'a
+    {
         &mut *(self.as_mut_ptr().offset((index[0] * self.row_stride() + index[1]) as isize))
     }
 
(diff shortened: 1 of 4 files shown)


tafia:
Forgot to mention that I had to change get_unchecked(_mut) function to make it work (I believe this particular change should actually go in master anyway)

src/matrix/slice.rs
```diff
@@ -66,7 +66,9 @@ pub trait BaseMatrix<T>: Sized {
     }
 
     /// Get a reference to a point in the matrix without bounds checking.
-    unsafe fn get_unchecked(&self, index: [usize; 2]) -> &T {
+    unsafe fn get_unchecked<'a>(&self, index: [usize; 2]) -> &'a T 
+        where T: 'a
+    {
```
Andlon: I don't understand what the explicit lifetime parameter here does. Would you mind explaining it to me?

Could it be that in the first case, the compiler automatically thinks that the reference has the lifetime of the struct, but in the second case you tell it that it explicitly has a different lifetime...? I'm confused, and very curious!

tafia:
Yes that's exactly that. This is what is done for slices in particular.

Andlon:
Very nice!

I hacked up some benchmarks based on your code:

```
test linalg::iter::mat_diag_iter_100_100           ... bench:         152 ns/iter (+/- 0)
test linalg::iter::mat_diag_iter_100_500           ... bench:         152 ns/iter (+/- 2)
test linalg::iter::mat_diag_iter_10_10             ... bench:          32 ns/iter (+/- 0)
test linalg::iter::mat_diag_iter_10_50             ... bench:          33 ns/iter (+/- 1)
test linalg::iter::mat_diag_iter_500_500           ... bench:         685 ns/iter (+/- 4)
test linalg::iter::mat_diag_manual_100_100         ... bench:         109 ns/iter (+/- 0)
test linalg::iter::mat_diag_manual_100_500         ... bench:         100 ns/iter (+/- 1)
test linalg::iter::mat_diag_manual_10_10           ... bench:          30 ns/iter (+/- 0)
test linalg::iter::mat_diag_manual_10_50           ... bench:          30 ns/iter (+/- 0)
test linalg::iter::mat_diag_manual_500_500         ... bench:         685 ns/iter (+/- 1)
```

As you can see, for larger matrices, they seem to perform almost exactly the same, but for whatever reason, in the 100x100 example the diagonal iterator performs a fair bit worse than manual. Not sure what's going on here!

tafia:
Yeah, I'm not an expert in profiling and I don't know what's happening either.

src/matrix/slice.rs
AtheMathmo: This is an unsafe change - I'm unsure exactly on the semantics but was discussing this on IRC. Using the lifetime like this is essentially giving the return type an unbounded lifetime. If you wanted to explicitly and safely give the lifetimes it would be like this:

```rust
unsafe fn get_unchecked<'a, 'b: 'a>(&'a self, index: [usize; 2]) -> &'b T { ... }
```

This implementation says that 'a and 'b may have different lifetimes, but 'b cannot outlive 'a. If the lifetimes are set like this you will find that this once again fails to compile.

_Note that I may have got the bound round the wrong way. Whichever way says that 'a lives longer than 'b is the correct order._

AtheMathmo:
From my understanding this is an unsafe change - as detailed in my other comment.

To give a slightly higher level reasoning. We _cannot_ safely have an iterator which can return more than one mutable reference to the same structure. The reasoning is that rustc has no way to determine that the pointers are indeed to different locations - we could `collect` our iterator and have break borrowing rules. This is why we **need** unsafe code and pointers to do this. We have to manually enforce the borrow rules ourself by ensuring that each of the returns `&mut T` references are to unique items.

Andlon:
> From my understanding this is an unsafe change - as detailed in my other comment.
> 
> To give a slightly higher level reasoning. We cannot safely have an iterator which can return more than one mutable reference to the same structure. The reasoning is that rustc has no way to determine that the pointers are indeed to different locations - we could collect our iterator and this would break borrowing rules. This is why we need unsafe code and pointers to do this. We have to manually enforce the borrow rules ourself by ensuring that each of the returned &mut T references are to unique items.

I'm slightly confused about this. That it needs to be unsafe is clear, but I don't understand why it needs to be implemented with pointers. In either case we are manually enforcing the borrow rules. In some sense, the two approaches are equivalent in that way. Could you elaborate on what the difference is...?

src/matrix/slice.rs
Andlon: Thanks for clarifying, @AtheMathmo! That makes sense. Guess we're back to pointers once again ;-)

AtheMathmo:
An easier way to convince ourselves this change is unsafe:

```rust
fn next(&mut self) -> Option<Self::Item> {
    unsafe { Some(self.square_matrix.$get_unchecked([0, 0])) }
}
```

This code change will still compile and would let us collect an iterator that has multiple mutable references to the same point: `diag_iter.take(5).collect::<Vec<_>>()`.

Our users could also do this outside of this iterator using just the `get_unchecked_mut` function.

src/matrix/slice.rs
AtheMathmo: > Guess we're back to pointers once again ;-)

Sadly yes - and maybe `mem::transmuting` just for luck too!

tafia:
It took me some time to digest. 

Sorry for the pull request and thanks for the review! 

I still don't fully understand why we are "allowed" to do some stuff with unsafe pointers but not with other unsafe methods.

AtheMathmo:
> It took me some time to digest.

It may have taken longer as I'm still trying to figure out some of the reasoning myself...

> Sorry for the pull request and thanks for the review!

There is no need to apologise! I should be thanking you - it is always good to discuss these things and I'd hate to miss out on any other suggestions in future :)

> I still don't fully understand why we are "allowed" to do some stuff with unsafe pointers

I think the only good answer to this is that raw pointers are nasty. Once we enter raw pointer land we lose all information about lifetimes and origin of the data. That is why dereferencing the pointers is unsafe. Because of this whenever we interface over raw pointers we have to be careful that we enforce all of the various borrowing and lifetimes rules ourselves.

AtheMathmo:
@tafia Am I ok to close this now?

Andlon:
> Because of this whenever we interface over raw pointers we have to be careful that we enforce all of the various borrowing and lifetimes rules ourselves.

I think this means that we need a way to test for lifetimes. Not saying we need to figure that out for the case of diagonal iterators, but this is something that can very subtly break in future changes without anyone noticing.

tafia:
Let me iterate once more, I'll remove the unsafe get_unchecked
modification.

I think having the range instead of the pos_end is simpler to understand.

AtheMathmo:
@Andlon I think that perhaps this is another case where the compilefail crate would be handy. I'm not sure how else to test this other than writing a bunch of `not-allowed` things and ensuring they remain `not-allowed`.

I agree it is very valuable though - much of the `MatrixSlice` work is _very_ unsafe. We may well find a few errors when writing these tests...

AtheMathmo:
> I think having the range instead of the pos_end is simpler to understand.

I think so too - hopefully this is not the source of the performance regression.

Andlon:
@AtheMathmo: yeah, we'll definitely need something like `compiletest` at some point. Still haven't got any response to the issue I posted there though.

tafia:
Reverted to a safe state. Unfortunately the perf regression is still here. I tried using a simple counter instead if a range iterator but it didn't improve. If someone has an idea of where it can come from I'd be very interested.

AtheMathmo:
This build is failing as we need to merge the latest master in - there is a conflict with newer version of the num-crate (we needed a feature before it was released there).

Also I'm of the opinion that we should go for performance even if the implementation is a little more complex. I'm not really sure why the regression is there and haven't been able to tweak it out - I have a few more ideas but if none work I think we should stick with the existing version.

(Although I agree with your change to the benchmarks - it is a fairer comparison)

AtheMathmo:
Although this is a much nicer implementation I'm closing due to the performance regression. Hopefully in future we can find a way to make this work!

PR #66, opened 2016-10-08
Refactor decompositions

Andlon:
As discussed in #40 (more precisely [here](https://github.com/AtheMathmo/rulinalg/issues/40#issuecomment-251081064)), I've reorganized the decompositions by moving individual decompositions into separate files. No functionality has been changed, it is merely for the sanity of the developers, as the `decomposition.rs` file was growing very large.

This opens up starting to overhaul the API for individual decompositions, as discussed in #40 (though we still have to agree on details).


AtheMathmo:
Looks good! It looks like a few files don't have new lines at the end - but I really don't mind and can merge as-is.

I'll also close #44 - for the most part this encompasses that, the only remaining differences are currently disputed :)

Andlon:
Ah, sorry about the line endings. That was sloppy of me. Let me fix it right away.

Andlon:
Line endings fixed. I'm OK to merge if you think it looks all right.

AtheMathmo:
Thanks for this - the code is a lot more manageable now!

Andlon:
@AtheMathmo: I thought I'd try to overhaul LU next, and see how it looks like with a custom struct. Should give us something more tangible to work with. However, I noticed some of these decompositions could definitely need some more tests while I'm in the business of overhauling, and so I think I'll turn my attention to introducing the `approx` crate first. While it provides some very useful macros for relative and ULP comparison, it actually currently doesn't have an absolute error macro, so I might have to throw in a PR [there too](https://github.com/brendanzab/approx/issues/8), first.

Hmm, I might want to write some custom matrix assertion macros anyway, as then we can provide much more specialized errors when they fail. Something like

```
397 out of 400 matrix elements compared equal.
The following elements did not meet comparison criteria:
(15,  3) - expected: 103.151351, actual: -99.001141
(14,  2) - expected:  13.000000, actual:  12.500000
( 3,  3) - expected:   0.000000, actual: - 3.153135
```

Might make an issue for this later on, if you think it could be useful. Think it would make debugging/development much easier though, and it would also be useful to our users.

AtheMathmo:
We should perhaps move some of this discussion to another issue later. But I'll respond here for now.

> I thought I'd try to overhaul LU next

This does some like a sensible next choice. It should be fairly straight forward if you save the more efficient storage stuff for later. (Though that is a cool benefit :) )

> However, I noticed some of these decompositions could definitely need some more tests

Yes! They certainly can do with more tests - especially the helper functions like householder/givens transformations. It may also help to introduce [quickcheck](https://github.com/BurntSushi/quickcheck) here but I'm not sure how much work this would be.

> I think I'll turn my attention to introducing the `approx` crate first

Truthfully I'm not too sure of the value here. To be clear, this is due to my own ignorance. I don't really have any experience with using ULP comparison etc. If you are convinced it is a helpful addition for testing I have no real objection. Would this be used only for testing or would you expect it to go into the main library too? (The first is an easier sell but I'm not totally against the latter).

> Hmm, I might want to write some custom matrix assertion macros anyway

I think these could be really valuable too. And we may as well add them for `Vector` too while we're at it?

Andlon:
@AtheMathmo: Agree with using quickcheck. I want to start using this very soon, but we can make the transition easier on ourselves if we give ourselves more tools to test with first (i.e. matrix macros etc., see #67).

> Truthfully I'm not too sure of the value here. To be clear, this is due to my own ignorance. I don't really have any experience with using ULP comparison etc. If you are convinced it is a helpful addition for testing I have no real objection. Would this be used only for testing or would you expect it to go into the main library too? (The first is an easier sell but I'm not totally against the latter).

(I'll just answer this here for now)

The problem with comparisons based on absolute equality is that they are completely unreliable unless you already know what data you are testing with. For example, consider two matrices:

```
// Here `I` denotes the identity matrix for some arbitrary dimension
A = 1e-13 * I
B = 1e-11 * I
```

The diagonal elements of B are 100 times greater than those of A, but if you compare them with an absolute tolerance of `1e-10`, they would happily compare equal.

Instead, you can compare them with relative error (i.e. `|b - a| / max(|a|, |b|) < r` for some r < 1). This works much better in the general case, but it's not so intuitive to pick `r`, at least in my opinion. The "last" alternative is comparing [Units in Last Place](https://en.wikipedia.org/wiki/Unit_in_the_last_place). I'm not overly familiar with the details here, but my impression is that this is typically used when you want to verify that two numbers are supposed to be exactly the same, but there are some very minor differences caused by floating point imprecision. ULPs work much better as a "default almost-exact" comparison for floating point numbers than absolute difference.

AtheMathmo:
Thank you! This definitely helps to make sense of things.

PR #68, opened 2016-10-15
Add assertions to cholesky doctest

andrewcsmith:
Verifies the structure of the return value, and verifies that we do not
unintentionally return an Error from valid data by unwrapping the result of
cholesky().

### DIFF
--- a/src/matrix/decomposition/cholesky.rs
+++ b/src/matrix/decomposition/cholesky.rs
@@ -15,11 +15,15 @@ impl<T> Matrix<T>
     /// # Examples
     ///
     /// ```
-    /// use rulinalg::matrix::Matrix;
-    ///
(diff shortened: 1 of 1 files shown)


AtheMathmo:
I believe these are already tested in the [regression tests](https://github.com/AtheMathmo/rulinalg/blob/master/tests/mat/mod.rs#L32).

To me it makes more sense to expand the tests there instead of the doc tests. Doc tests should be as simple as possible whilst showing usage (though this isn't the case everywhere...).

andrewcsmith:
I see what you mean. I do prefer to have documentation be explicit about what is being returned (thinking of the convention of using ` #=> ...` in Ruby documentation) but if you want to keep things separate that's find by me.

PR #69, opened 2016-10-15
Matrix authoring docs

bright-star:
Since Rust is a systems language and therefore the "how" matters, here's a PR I've started to add a big tutorial section on authoring matrices. I need help to fill in all the sections, but I wanted to get started. Includes features that will be added in #55, since I'm already testing them locally.

### DIFF
--- a/src/matrix/mod.rs
+++ b/src/matrix/mod.rs
@@ -21,6 +21,8 @@
 //! zeros:
 //!
 //! ```
+//! use rulinalg::matrix::Matrix;
+//!
 //! let mut my__zero_mat = Matrix::<f64>::zeros(10, 20);
 //! ```
 //!
@@ -34,6 +36,9 @@
 //! elsewhere:
 //!
 //! ```
+//! use rulinalg::matrix::Matrix;
+//! use rulinalg::vector::Vector;
+//!
 //! let main_diag = Vector::<f64>::ones(5).apply(&|el| { el * -2 });
 //! let mut diag_mat = Matrix::from_diag(main_diag.data());
 //! ```
@@ -50,6 +55,9 @@
 //! be 1:
 //!
 //! ```
+//! use rulinalg::matrix::{Matrix, DiagOffset};
+//! use rulinalg::vector::Vector;
+//!
 //! let main_diag = Vector::<f64>::ones(5).apply(&|el| { el * -2 });
 //! let mut diag_mat = Matrix::from_diag(main_diag.data());
 //! for el in diag_mat.iter_diag_mut(DiagOffset::Above(1)) { *el = 1.0; }
--- a/src/matrix/mod.rs
+++ b/src/matrix/mod.rs
@@ -39,7 +39,7 @@
 //! use rulinalg::matrix::Matrix;
 //! use rulinalg::vector::Vector;
 //!
-//! let main_diag = Vector::<f64>::ones(5).apply(&|el| { el * -2 });
+//! let main_diag = Vector::<f64>::ones(5).apply(&|el| { el * -2.0 });
 //! let mut diag_mat = Matrix::from_diag(main_diag.data());
 //! ```
 //!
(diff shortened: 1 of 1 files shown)


Andlon:
Thanks for putting in the effort to improve the docs @bright-star! Like all libraries in its infancy, I guess the docs tend to be a bit neglected (at least for now). My only concern with tutorial-type documentation at this point is that it can easily get out-of-sync, since I believe the API of `rulinalg` is still very much in flux. Doctests help, of course, but it might become an effort to make sure the tutorial showcases the best ways of working with the library at any given time.

Looking over your contribution, I have the impression that most of the examples are relatively isolated with respect to individual methods. By that I mean that each of your headers essentially deal with a single primary method on matrices, such as `from_diag` or `diag_iter_mut`. Could it be an idea to instead use the effort here to improve the examples for the individual methods, while instead using the centralized documentation to link directly to methods that are relevant for the construction of matrices? (Effectively making it an index of such methods) I don't know if this is better or not - it's really just an idea - but I'd like to hear your opinion!

bright-star:
@Andlon I agree with what you're saying in principle, and you're definitely right that I need to avoid redundantly indexing methods.

On the other hand, a tutorial section meets a legitimate need: I started trying to port some code to use this library, and I couldn't tell from the API docs which methods I was supposed to be using. API docs, like unit tests, only answer the "what": what methods are available. They don't tell you "which": which method to use for what purpose.

Here's an example: if I want to modify the first row of a Matrix, should I use `.get_row_mut(0)` or `.iter_rows_mut().take(1)` or `.sub_slice_mut([0,0],  0, &mat.cols()-1)`? Or is there something inside Vector that I should look at? I don't know (truthfully), because the API docs, by nature, can't tell me which.

I don't mind reorganizing it as you say, but I want to make sure that this much is covered, so people can "just start" using the library.

src/matrix/mod.rs
```diff
@@ -5,6 +5,71 @@
...
+//! If the main defining feature of the matrix is its diagonal, you can create a
+//! vector with the appropriate values and use the `from_diag` method. For
+//! example, here is a 5x5 matrix with -2 on its main diagonal and zeros
+//! elsewhere:
+//!
+//! ```
+//! use rulinalg::matrix::Matrix;
+//! use rulinalg::vector::Vector;
+//!
+//! let main_diag = Vector::<f64>::ones(5).apply(&|el| { el * -2.0 });
```
AtheMathmo: I think using `Vector` here might be a little confusing, I would stick with just `Vec`.

AtheMathmo:
I just wanted to chime in and say that I do think these are very valuable to have. I do think what is missing from the API is some explanation of when each type of construction and various other functions are best used. This is one of the things I most frequently address when taking contributions to rusty-machine, because it is complex and not explained very well.

I'm not sure how much this PR can fix there, but certainly I'm happy to see any movement in that direction.

Andlon:
@bright-star: thanks for the perspective! I see your point, and I agree that a user guide of sorts has a place in the documentation.

I've given this a little thought, and I think a very important point to make is that you can always resort to simply indexing into your matrix, i.e. `mat[[i, j]]`, which is intuitive, but not always the most convenient. Everything else is basically more convenient and/or more performant (in the sense of elision of bounds checks) accessors. Perhaps you could lead with this story, so that you immediately give users the tools to form _any_ kind of matrix (by working directly with the indices), and then later show how `rulinalg` has functionality for more convenient manipulation of special patterns in matrices? (row/col/diag iterators, slices etc.) Just an idea!

bright-star:
@Andlon Got it. What I'll do for now is finish the port I was doing using the most blunt and simple tools, then extract simple examples from it and place them here, so you can examine them.

AtheMathmo:
I'm going to close this due to inactivity (it frees up a little mental space for me to focus the little time I have).

Feel free to reopen if you want to continue adding this to the docs. I do think it could be a valuable resource.

bright-star:
No problem. I think the PRs/issues about special matrices come first.

Issue #67, opened 2016-10-08
Matrix equality macro


Andlon:
Currently some of our tests look like this:

``` rust
assert!(!mat.data()
    .iter()
    .zip(recovered.data().iter())
    .any(|(&x, &y)| (x - y).abs() > 1e-10));
```

This is functional, and it _works_, but it has some issues:
- It is not immediately obvious what it does, and as such subtle errors may easily creep in.
- It is verbose.
- If the assertion fails, it's very hard/inconvenient to determine what actually went wrong, because it is boolean.
- Most of our tests currently use arbitrary thresholds (*)

It would be immensely useful for us, if we instead could write something along the lines of

``` rust
// Elementwise absolute difference comparison
assert_matrix_eq!(mat, recovered, tol = 1e-10, kind = absolute);

// Frobenius norm comparison, perhaps?
assert_matrix_eq!(mat, recovered, tol = 1e-10, kind = frobenius);

// Default comparison (X amount of Units in Last Place (ULP) perhaps?)
assert_matrix_eq!(mat, recovered);
```

With such a macro, we could have much more detailed feedback when the assertion fails. For example, for a small matrix:

```
Expected:
1.00000, 0.00000, 0.00000,
0.00000, 1.00000, 0.00000,
0.00000, 0.00000, 1.00000

Actual:
1.00000, 0.00000, 0.00000,
0.00031, 1.00001, 0.00002,
0.00000, 0.00312, 1.00812
```

or for a larger matrix (bigger than 5x5 or so?)

```
397 out of 400 matrix elements compared equal.
The following elements did not meet comparison criteria:
(15,  3) - expected: 103.151351, actual: -99.001141
(14,  2) - expected:  13.000000, actual:  12.500000
( 3,  3) - expected:   0.000000, actual: - 3.153135
```

While this is immensely useful for the developers of `rulinalg`, keep in mind that our users also need to write test cases for their applications, and as such it would be very useful also for them.

Note that I have not considered the above very thoroughly. I rather wanted to throw out some ideas, which might inspire other good ideas.

I'd like to start experimenting with this and see what I can come up with, when I have time.

(*) On arbitrary thresholds: this obviously isn't fixed by a macro. Instead, at some point we need to try to design our tests to use theoretically proven bounds for errors. This is sort of next level though, but it has to be done if `rulinalg` wants to be a serious alternative in the longer run.

Andlon:
[...]

PR #71, opened 2016-10-16
Matrix equality assertion

Andlon:
This pull request is not intended as a candidate for merge (yet).

I'm working on the matrix equality macro mentioned in #67. I added an example in the `examples/` subfolder, which one can run with `cargo run --example mismatched_f64_elements`. See also the small number of tests currently included as well. The example looks like this:

``` rust
#[macro_use]
extern crate rulinalg;

fn main() {
    // The purpose of this example is to demonstrate the type of errors generated
    // by `assert_matrix_eq!` when the comparison fails (which the below call will).
    let a = matrix![1.00, 2.00;
                    3.00, 4.00];
    let b = matrix![1.01, 2.00;
                    3.40, 4.00];
    assert_matrix_eq!(a, b, comp = abs, tol = 1e-8);
}
```

and it gives the following output:

```
     Running `target/debug/examples/mismatched_f64_elements`
thread 'main' panicked at '

Matrices X and Y have 2 mismatched element pairs. The mismatched elements are listed below, in the format
(row, col): x = x[[row, col]], y = y[[row, col]].

 (0, 0): x = 1, y = 1.01. Error: 0.010000000000000009
 (1, 0): x = 3, y = 3.4. Error: 0.3999999999999999

Comparison criterion: absolute difference, defined by
    |x - y| <= 0.00000001.

', examples/mismatched_f64_elements.rs:11
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

Now, I'd like to give people the chance to give some input on how the macro should work. Comparing floating numbers is _seriously hard_. See for example [this famous article](https://randomascii.wordpress.com/2012/02/25/comparing-floating-point-numbers-2012-edition/). A preliminary conclusion is that there is no one-size-fits-all solution for comparing floating point numbers. Hence, it's impossible for us to have a default `assert_matrix_eq!(a, b)` macro that works in all cases. This is why I've opted for explicitly making the user choose which criterion to assert with (the `comp = abs` and `tol = 1e-8` parameters).

However, I think we can provide a very conservative default `assert_matrix_eq!(a, b)` macro. I need to think a bit more on exactly how it's going to work, but I have the following in mind:
1. The default invocation is very strict and conservative.
2. The user simply writes `assert_matrix_eq!(a, b)` for his/her tests, and this might be fine if the numbers are very close to exactly equal.
3. However, in the case when the numbers are near but not sufficiently near, the macro will fail.
4. When the macro fails, the error message should explain the next steps the user should take (ideally by referring to a fleshed out part in the documentation on this) to ensure that he/she is using the correct comparison criteria.

The reasoning in step 4 again comes back to the fact that a _one-size-fits-all comparison does not exist_. I'm a bit inspired by how `rustc` guides users around compilation failures, by giving error messages that are truly helpful. Hopefully we can do the same for our users (and for our contributors).

Currently, the macro only supports absolute difference as a comparison criterion, but I want to at least support [ULP](https://en.wikipedia.org/wiki/Unit_in_the_last_place)-based comparison too (probably basing the default criterion on something involving ULP). I also need to write a **lot** more tests, as I hope this could form the cornerstone of future test assertions in rulinalg as a whole, as well as flesh out the documentation much more. Also, there should perhaps be an analogous `assert_vector_eq` macro?

Would really appreciate any opinions, questions, reservations or advice anyone might have!


src/macros/matrix_eq.rs
```diff
@@ -0,0 +1,221 @@
...
+                // TODO: Aligned output
+                let mut formatted_mismatches = String::new();
+                for mismatch in mismatches {
+                    formatted_mismatches.push_str(" ");
+                    formatted_mismatches.push_str(&mismatch.to_string());
+                    formatted_mismatches.push_str("\n");
+                }
+                Some(format!("\n
+Matrices X and Y have {num} mismatched element pairs. The mismatched elements are listed below, in the format
+(row, col): x = x[[row, col]], y = y[[row, col]].
```
AtheMathmo: Super minor point - I think this should read:

`(row, col): x = X[[row,col]], y = Y[[row, col]].`

examples/mismatched_f64_elements.rs
```diff
@@ -0,0 +1,12 @@
+#[macro_use]
+extern crate rulinalg;
+
+fn main() {
```
AtheMathmo: I think it is valuable for the user to see the example output (perhaps annotated) without having to run the example. I mean something like what we have in [rusty-machine](https://github.com/AtheMathmo/rusty-machine/tree/master/examples).

If you think adding a README to the examples doc is overkill, perhaps we could just add a block comment at the end of the main function detailing the output?

src/macros/matrix_eq.rs
Andlon: Indeed it should!

examples/mismatched_f64_elements.rs
Andlon: I was thinking perhaps to cover these things in a more detailed way in the documentation itself? I.e. in the documentation for `assert_matrix_eq!`, which would then be complete with examples and their outputs. The runnable example is as much for us as developers as for the users, I guess. Without it, there's no quick way to manually verify the output message of the macro.

src/macros/matrix_eq.rs
```diff
@@ -0,0 +1,221 @@
...
+            }
+        }
+    }
+
+    fn description(&self) -> String {
+        format!("absolute difference")
+    }
+
+    fn definition(&self) -> String {
+        format!("|x - y| <= {tol}", tol = self.tol)
```
AtheMathmo: I was looking into whether we can use scientific notation here for readability. (`"{tol:e}"`).

Looks like it is only implemented for `f32` and `f64` by default. In which case we should leave it as-is?

AtheMathmo:
This PR is really great! I really like how the macro works and the output is pretty much perfect (I think).

A few thoughts:

- Should this macro capture `a == b` as well? I think not but wanted to put it out there as a potential comparison function.
- The default implementation ideas you have sound really good. If this turns out to be a lot of work I think something less sophisticated would be acceptable. Maybe just default to abs and a low tolerance - but add a note in the panic message that the user can change these settings.
- As you say I think we should add an `assert_vector_eq` macro too.

How do you plan to implement ULP?

Thanks for this!

src/macros/matrix_eq.rs
```diff
@@ -0,0 +1,221 @@
...
+/// let a = matrix![1.000, 2.000,
+///                 3.000, 4.000];
+/// let b = matrix![0.999, 2.001,
+///                 2.998, 4.000 ];
+/// assert_matrix_eq!(a, b, comp = abs, tol = 0.01);
+/// # }
+/// ```
+#[macro_export]
+macro_rules! assert_matrix_eq {
+    ($x:expr, $y:expr, comp = abs, tol = $tol:expr) => {
```
Andlon: Yes, the intention is to allow specifying different criterion. I'm not so sure of the syntax yet, but if we keep the current one, I can imagine something like

```rust
assert_matrix_eq!(x, y, comp = ulp, tol = 4)
```

to assert within 4 ULP. However, I'm not so sure if we even need to expose an ULP based comparison to the user, or if we just want to use it internally for the default comparison. In any case, I strongly think that some form of "comp = abs" (possibly using a very different syntax) should be present in the macro invocation, so that it's completely clear what sort of comparison is being used (implicit assumptions about floating point comparison can lead to all kinds of subtle and hard-to-track-down bugs).

It might also be useful later if we want to add comparators that are not elementwise, such as based on matrix norms, for example. This might be useful if you know that an algorithm has an error measured by some norm, i.e. ||A-expected|| <= tol.

src/macros/matrix_eq.rs
Andlon: Agree. I want to spend some time working on the quality of the output (alignment and precision etc.) once the more high-level features are in place. Scientific notation would definitely be useful.

Andlon:
@AtheMathmo,

I think it makes sense to introduce "comp = exact", perhaps? (with no tolerance) There are actually cases when you want to compare floating point numbers exactly, although you usually don't. Moreover, for integer types or boolean types, this certainly makes a lot of sense.

I have to figure out exactly how I'm going to support all the different types (`bool`, unsigned and signed integers, floating point at least... Complex numbers? Might leave complex numbers for a separate issue) in a decent manner. Ideally you want to be able to use the default macro with no additional arguments for all of them, which may require some sort of "specialization" through trait bounds or so, since the behavior for e.g. bools and floats must be different.

I also agree with the notion that I should not spend _too_ much time on this, as it detracts from the time I could spend on improving arguably more important parts of the library. However, I think if I manage the number of features the macro should have in a sensible way, it makes sense to spend some time on making sure this works properly, in which case it should make future testing easier.

For ULP, I was initially planning to use the `approx` crate, however I realized that since it only gives you a boolean (i.e. "this is sufficiently close, or it is NOT sufficiently close within some ULP tolerance"), it doesn't give me enough information to give valuable feedback to the user. For example, I'd like to give the error in ULPs as well (corresponding to the "Error: 0.399..." in the output, but in ULPs instead). I looked at the implementation in `approx`, and it's fairly straightforward. I might use a similar approach but tweak it a little to better support our use case.

As I mentioned in one of the review comments though, I'm not sure if we actually want to expose ULPs to the user or just use it internally for the default macro invocation.

Andlon:
On booleans: never mind that. At the moment, there is not enough support in `rulinalg` to warrant using a `Matrix<bool>`, and in that case you might just use `Matrix<u8>` or so instead. In any case, I think I'll leave bool out of it for now (as it potentially complicates the macro implementation quite a bit).

Andlon:
Just leaving a note for myself here, so I remember it. Need to put some maximum bounds of number of element pairs to show errors for (i.e. 10-15 or so), so that we don't accidentally print 10000 lines of deviating element pairs :smile:

AtheMathmo:
Good thought!

Maybe showing the first 5-10 and then saying how many there are in total?

Date: Mon, 17 Oct 2016 05:33:59 -0700

AtheMathmo:
I was just wondering how much more you wanted to do on this? With the last couple of weeks I've lost track a little, sorry!

Andlon:
Just want to give an update on this. My current problem is that I have not found a way to reliably compute ULP differences given only a `Float` trait bound. I'd rather avoid having to implement a custom trait for floating point numbers, as it means exposing it for our users (in case they have their own floating point types).

There is `Float::integer_decode()`, but without knowing the actual width of each field, I can't think of a way to reliably compute the ULP difference. For example, I can use i.e. `Float::max_value()` to deduce this information, but only if I assume a IEEE754-type format, which the `Float` trait at least does not seem to guarantee, and it really sounds like a hack in any case.

I could also forego ULP for now, but providing only absolute difference is not so valuable in the general case. I could implement relative differences, but they are really not so intuitive to use. I need a bit more time to think on this, I think.

Andlon:
Actually, I think the custom trait way is the only reasonable way at the moment (short of sending pull requests for breaking changes into `Num`). That is, to provide our own `Ulp` trait, which would provide a `ulp_diff()` function and which we implement for `f32`, `f64`, and document how to implement it for custom types. Is this an acceptable solution to you?

AtheMathmo:
I think the custom trait is a fine approach. It feels as though rulinalg isn't the natural place for such a trait but for now at least I see no issue with it.

I think it also worth opening an issue for discussion on the `num` crate. It would be good to at least find out if they would consider adding something like this. It wouldn't have to be a breaking change if it were introduced as a standalone trait initially.

Andlon:
@AtheMathmo,

all right! I'll look into it as soon as I have time. ULP does not fit perfectly into the current approach, as ULP does not make sense for certain situations (such as numbers of different signs or NaNs).

As for `num`, I think once we've landed on something here, we can consider opening an issue for `num`. I think in particular information about the size of the fields in `integer_decode` would be sufficient, and I think that could also make sense for them to include.

Andlon:
OK, so the three main classes of comparators are now implemented:

- Exact comparison
- Absolute difference comparison
- ULP-based comparison

Apart from more documentation, much more and better tests and improved output (alignment, scientific notation, limit on number of disparities), the main feature that is lacking is a _default_ comparison when you just call `assert_matrix_eq!(x, y)`. Ideally I would want the following behavior, which is inspired by the `AlmostEqualUlpsAndAbs` function in [this article](https://randomascii.wordpress.com/2012/02/25/comparing-floating-point-numbers-2012-edition/):

1. If the scalar type is a floating point type:
    1. If `|x - y| <= C * epsilon`, return true. Here C is a small constant, between 1 and 4 or so.
    2. If `ulp_diff(x, y) <= 4`, return true, otherwise false. 4 somewhat arbitrarily chosen.
2. Else return true if `x == y`, otherwise false.

I'm not entirely sure how to get this type-conditional behavior though. I'll think about it until I have some more time to continue the implementation. In the meantime, do you have an idea, @AtheMathmo? (or anyone else, of course!)

AtheMathmo:
I can't think of a clean way to get this kind of behaviour. I think it would involve creating a new trait for the default comparison, `DefaultComparison` or something. And for float primitives we do the comparison 1, and for all others comparison 2. The issue is that this would have to be public facing so that users can implement them on their own types. I'm also not sure if this would even fit well into the macro? I really haven't played with macros too much.

Andlon:
Yes, something along those lines is what I have been considering. Is it possible to use generic implementations with trait bounds to get the effect? I.e. we'd implement `DefaultComparison for T where T: Float + Ulp` and `DefaultComparison for T: PartialEq<T>` or so. Then it would stricly speaking have to be a public trait, but users do not need to implement it, and we can hide the docs. I'll try this out next time I work on it, but I'm not sure if Rust allows such overlapping trait bounds?

AtheMathmo:
Sadly Rust does not allow this kind of generic trait implementaiton - I'm sure for good reason. This is the same reason why it is so difficult to implement lhs addition/subtraction operators (`impl<T> Mul<Matrix<T>> for T where ... {}`)..

We'd have to do some macro magic to cover the primitives, and in the rare case that a user needs to implement this trait they'd be on their own I suppose.

Andlon:
I see. I suppose this is related to the lack of specialization?

In that case, I'm not sure if the required macro magic is worth the effort. I have two different suggestions:

1. The default `assert_matrix_eq!(x, y)` will only be valid for floating point types, and we require the trait bounds `Float + Ulp`. In this case we can use the absolute epsilon + ulp comparison I talked about earlier.
For other types, i.e. integers or whatever, the user explicitly needs to use `assert_matrix_eq!(x, y, comp = exact)` or similar. The downside here is that it may not be so clear to the user why the default comparison fails for i.e. an integer matrix.
2. The default trait uses exact comparison, and we can introduce an additional `comp = float` to use the combination of epsilon abs + ulp which I mentioned above. When the exact comparison fails, we can refer the user to the documentation for how to handle floating point types or other types of comparison in the error message.

Do any of these suggestions sound workable to you?

AtheMathmo:
I think I prefer the sound of your second suggestion. I think if I happened on a new library I would expect an `assert_eq`. Having access to a nice default for floats (which presumably is well documented) should cover all of our bases.

Andlon:
@AtheMathmo, I completely agree. I also much prefer the second option, just wanted to get your thoughts on the matter. I'll try to implement it next time I sit down with this!

Andlon:
By the way, I just want to explain why I'm so insistent on having such a macro for `rulinalg`, and why I'm prioritizing it over other matters, such as the decomposition redesign.

I've recently been building some C++ code with Eigen for my thesis, and the lack of similar facilities for testing is a **monumental** pain in the ass. I rely heavily on unit tests and property-based testing (RapidCheck for C++), and especially debugging these tests when they fail has sometimes been very frustrating. My hope is that my providing a useful macro out of the box, we can prevent both ourselves and the users of `rulinalg` from similar pains.

AtheMathmo:
I'm really happy that you are! I totally agree that it is necessary and hopefully an area where we can have good coverage from the start. It's going to make my life alone much, **much** easier.

If there is anything else that you think would have helped (or _did_ help) with your current work please feel free to write up any related issues. My personal use of rulinalg is [fairly narrow](https://github.com/AtheMathmo/rusty-machine) but I want to make it valuable to as many people as possible!

AtheMathmo:
I absolutely don't want to derail you, but I've been meaning to write this up for a while and thought it might be interesting to you/relate to this work: #81 .

Perhaps later down the line this could provide some more functionality to this macro.

Andlon:
@AtheMathmo: yes, this is part of the reason I chose the current syntax, as it allows extension in the form of, i.e.:

```rust
// I just made the syntax here up on the spot, so might not be optimal
assert_matrix_eq!(x, y, metric = LNorm::Infinity, tol = 1e-6)
```

AtheMathmo:
Had a quick look through, I think things are looking really good. The only issue I had was that I think the examples are a little unwieldy. I'm not against merging as-is, but would like to suggest the following ideas:

- We move these example into a new folder: `examples/macros/*` or `examples/equality/*`?
- We combine these into one example where the comparator is defined by a command line argument? So usage would look like: `cargo run --example equality -- exact`. And if no argument is given we use the default?

I prefer the second option. Let me know what you think.

Andlon:
@AtheMathmo: to tell you the truth, I made the examples mostly to guide my own development (i.e. so I can quickly see how the error messages look like in various scenarios). I think your suggestion is really good, but I would also be up for just outright removing the examples when I'm done.

I hope to have some time this weekend to work on it. I started on some more comprehensive documentation yesterday, but it's going to take me some time to test this properly. Since this is intended for use as the backbone of our future testing, it would be unfortunate if it contains bugs (although some might still slip through, of course).

This might actually be a good opportunity to try out QuickCheck. I recently started using the C++ library RapidCheck for some C++ code I'm developing, and it let me discover some truly obscure bugs in my mesh refinement algorithms that none of my (fairly detailed) unit tests covered. Would you mind if I add QuickCheck as a test dependency?

AtheMathmo:
Yes I'd be really happy to see quickcheck added. I want to use it elsewhere too so fire away!

I'll leave the examples up to you. If adding a cmd line argument is too much effort I'd be fine with removing them completely. As you say the focus should be on making sure this is all correct :D

Andlon:
@AtheMathmo: one thing. I think it should be documented which traits bounds apply to the scalar types in the comparison. As it is, I've hidden the docs from the underlying types used in the macros because I don't really want to expose how it works. At the same time, this also hides the type information and consequently the trait bounds. I could of course explicitly write out the expected trait bounds. This seems like it could easily become out-of-sync. I'm not sure which is better: writing out the scalar types explicitly, or exposing the underlying types in the documentation? What's your opinion?

Andlon:
Also, I'm currently struggling with a complication. Because `assert_matrix_eq!(x, y)` calls a generic function in the body and passes `&x` and `&y`, it works fine if `x` and `y` are owned. However, if `x` and `y` are references, such as `let ref x = matrix![];`, then Rust cannot deref `&&x` to `&x` automatically, and the macro currently fails. I'm not sure how to solve this. 

One way is to let it work like any Rust function, which means we'd have to write `assert_matrix_eq!(&x, &y)` if `x` and `y` are owned. I guess this is not a very big deal, but it IS more noise than necessary. On the other hand, it's explicit, like Rust should be. Thoughts?

AtheMathmo:
Sorry for the late reply. I read this yesterday and couldn't figure out a good approach. Yours sounds fine to me - though I wonder if there is some way to have the macro accept both? For example the `println!` macro will accept wither `x` or `&x`.

Regardless I think your approach is totally fine for now and perhaps the best we can get.

Andlon:
@AtheMathmo: `println!` doesn't need to call generic functions, I guess. In any case, did you see my edit? I managed to find a workaround, so the macro actually works fine with both `x` and `&x` now.

AtheMathmo:
@Andlon sorry I wasn't very clear, yes I saw the edit and I think it probably is the best approach :)

Andlon:
Slowly getting there... Remaining work:

- [ ]  Some tests for `elementwise_matrix_comparison` to ascertain that it corrects the right elements as mismatching.
- [ ] Better formatting, i.e. scientific notation and some alignment.
- [ ] Need to implement analogous `assert_vector_eq!`
- [ ] Put an upper bound on number of mismatches reported (i.e. 10-15) to prevent flooding the terminal.
- [ ] Add examples of error reporting to documentation
- [ ] Figure out which course to take with the documentation trait bound problem mentioned a few comments up. @AtheMathmo: got any input here?
- [ ] \(Optional/not sure if the best idea) Test exact output of error reporting? (Otherwise we have no automated tests which check that the textual output is as expected)

List looks long, but each point is not so much work. I'm done with writing most of the tests, which I expect was the most time consuming part.

AtheMathmo:
@Andlon sorry I have been a little slow to respond here. Your list of remaining work looks good to me. I'm also unsure about testing the output - I feel like we will just have to update this test when we make any changes and it doesn't really lower the chance of any typos/bugs getting in.

As for the documentation issue, I'm not totally sure that I understand. But explicitly putting the trait bounds in the documentation sounds acceptable but not ideal. As you say I think it could get easily out of sync. It would be preferable if we could automatically show the required bounds from the compiler. If that isn't possible I think stating explicitly is fine. Personally I'd be inclined to reveal what's under the hood a little but I don't know exactly what you're concealing and your purpose for doing so. If you believe it is best hidden that may be the way to go.

Andlon:
About scientific notation. In order to use `{:e}` in `format!`, `T` needs to implement `std::fmt::LowerExp`. However, this is only implemented for `f32` and `f64`, but our `abs` comparator needs to work for both integers and floating-point numbers. Without trait specialization, I am not sure how to deal with it. Unless anyone has an idea, I suggest we just leave it with the default output for now, and if it turns out to be too burdensome, we make an issue on it later.

Another question is whether or not absolute-valued comparisons are useful for integers. I suppose usually you are only interested in an exact result in the first place? I can't even think of a situation where I'd be interested in comparing matrices of integers with anything but exact equality. We might perhaps consider reserving `abs` for floating-point numbers only?

AtheMathmo:
I think I saw that issue before, with `LowerExp`. It's a little odd that it isn't implemented for integers (though understandable). Let's leave it out for now then?

I agree that `abs` for integers seems a little unnecessary. If it is extra effort to include it let's leave it out too. If it is more effort to exclude it then we may as well keep it in?

Andlon:
While there are still things that could be improved upon, I consider this ready for review/merge now. Ideally the `abs` comparator would be constrained to `LowerExp` so we could get scientific notation (and consequently exclude integers from use), but I consider this a low priority feature, and I'd rather work on more pressing matters.

In the end, it turned out to be a rather large body of work. I spent a lot of time writing tests, but I find this necessary since the macros are intended to be used as core building blocks in writing future tests, and as such they arguably need to be the most reliable part of the library (though bugs may still have crept in).

I also suggest that we don't immediately rewrite existing tests using these macros, but rather encourage using it when writing new tests. In the long run, if we encounter no problems, we may perhaps start thinking about replacing our older tests.

@AtheMathmo: I'd like to start exploring some of my ideas for custom decomposition types (#40) next. I'd like to create a proposal for a redesign of the LU decomposition first. Do you think this is reasonable, or would you rather I work on something else? I'm up for suggestions. After the menial work of writing all these test cases for this macro, I need something more interesting, in any case, and I think the decomposition redesign would fit the bill.

Andlon:
Also, I experimented some with testing the output of `panic_message()`, but it was a nightmare to determine what was wrong if the test failed (i.e., there could be a space missing somewhere or similar). Hence, we currently have no tests to verify the actual _textual_ output of the macros, but their behavior is quite well tested in all other aspects.

AtheMathmo:
Glad to see this is ready for review! I'll try to find some time today to check it all over. The large number of tests makes me confident in this.

>  Ideally the `abs` comparator would be constrained to `LowerExp` so we could get scientific notation

I agree that this would probably be a nice change but can definitely wait a while.

> I also suggest that we don't immediately rewrite existing tests using these macros, but rather encourage using it when writing new tests.

This sounds like a good idea. I'm pretty excited to make use of these but we should hold off until time makes us a little more confident.

>  I'd like to create a proposal for a redesign of the LU decomposition first.

Awesome! I'm really excited for this work to begin and I agree that LU is a good start. From memory I think where we left the discussion before I was happy with the ideas we had. I look forward to reading the proposal and hopefully seeing the subsequent work!

> we currently have no tests to verify the actual textual output of the macros

This is absolutely fine. As I said above I'm not sure in practice that these would be the most valuable anyway. We'll just rely on users helping us to find any bugs in the output :).

Andlon:
FYI: I just added a small commit to the docs which details where the requisite trait bounds for matrix/vector elements can be found for each comparator.

src/macros/matrix_eq.rs
```diff
@@ -0,0 +1,1564 @@
...
+                    formatted_mismatches.push_str("\n");
+                }
+
+                // Strip off the last newline from the above
+                formatted_mismatches = formatted_mismatches.trim_right().to_string();
+
+                Some(format!("\n
+Vectors x and y have {num} mismatched element pairs.
+The mismatched elements are listed below, in the format
+#index: x = x[index], y = y[index].
```
AtheMathmo: I think this is a little odd looking, as both the vectors and the elements take the same letter in lower case. I'm not sure what the solution should be. Perhaps use `u` and `v` for the elements? Otherwise we can use upper case for the vectors (not ideal)?

We can probably just leave it as it stands. I think it's obvious enough what is meant.

src/macros/matrix_eq.rs
```diff
@@ -0,0 +1,1564 @@
...
+    Match,
+    MismatchedDimensions { dim_x: (usize, usize), dim_y: (usize, usize) },
+    MismatchedElements { comparator: C, mismatches: Vec<MatrixElementComparisonFailure<T, E>> }
+}
+
+/// Trait that describes elementwise comparators for [assert_matrix_eq!](macro.assert_matrix_eq!.html).
+///
+/// Usually you should not need to interface with this trait directly. It is a part of the documentation
+/// only so that the trait bounds for the comparators are made public.
+pub trait ElementwiseComparator<T, E> where T: Copy, E: ComparisonFailure {
```
Andlon: We don't need the `Copy`, but we do need at least a `Clone`, as I'm storing all of the mismatches. It _is_ possible to do it all with just references, but I suspect the code would be less manageable. I think for now we might want to keep it as is, and then perhaps consider improving this in the future. Currently I don't believe it is truly worth the effort.

src/macros/matrix_eq.rs
Andlon: Agree. I'd rather not want to introduce more symbols than necessary, so I suggest we either leave it as is or use capital X and Y. I don't mind either way, so I'll let you decide :)

src/macros/matrix_eq.rs
```diff
@@ -0,0 +1,1564 @@
...
+    } else {
+        VectorComparisonResult::MismatchedDimensions { dim_x: x.len(), dim_y: y.len() }
+    }
+}
+
+#[doc(hidden)]
+#[derive(Copy, Clone, Debug, PartialEq)]
+struct AbsoluteError<T>(pub T);
+
+/// The `abs` comparator used with [assert_matrix_eq!](macro.assert_matrix_eq!.html).
```
Andlon: Huh, strange. All the links worked fine for me... I'll have to test again and investigate this. Were you on 1.13 stable?

Andlon:
Thanks for the comments! Will try to push up the changes tonight, if I have the time.

Andlon:
Forgot to address the custom comparators issue. Well, actually, it's currently not actually possible for a user to implement their own custom comparator that works with the macro. It wouldn't take too much extra work to support this, but that would be a bit of an extension that I'm not sure if it's worth it yet. May I perhaps suggest we try out the macro and see how well it works in practice, and then if we see the need for this, we can create an issue for improvements to the macro?

src/macros/matrix_eq.rs
AtheMathmo: I just retried on 1.13 and saw the same issue. If I am on `.../rulinalg/macros/trait.ElementwiseComparator.html` and I click the link, I go to: `.../rulinalg/macros/macro.assert_matrix_eq!.html`.

Maybe this is only an issue on my system? Or the way I'm viewing the files (from my file system with chrome).

AtheMathmo:
Take your time with the changes :).

I agree that leaving custom comparators out for now is fine. Maybe you can write up an issue detailing how we do this? I imagine it will be a while until (if) we get around to it but having it there ready may be valuable.

One other _very_ minor thing that I forgot about before. Should the `ElementwiseComparator` `compare` function return a `Result` instead of an `Option`? I don't care much either way, but as we return an error maybe this is more natural?

Andlon:
Sure, I can write up an issue with remaining ideas for improvement after we've merged (if I somehow forget, feel free to remind me).

As for the `compare` function, I totally agree! It's a bit inverted as it is. Always felt wrong to me. I agree that a `Result<(), ComparisonFailure>` (pseudosyntax) would make more sense. Thanks :)

src/macros/matrix_eq.rs
Andlon: Ah! That makes sense. The link needs to be `/macro.assert_matrix_eq!.html`, I think? Will test it out next chance I get. Thanks for the feedback!

src/macros/matrix_eq.rs
Andlon: Capital letters it is then :)

Andlon:
@AtheMathmo: I believe I have addressed your concerns with the latest commits! (I hope)

Turns out links are rather fragile things in `rustdoc` at the moment. They should all work now, but if you move structs around between modules they may easily break. I think they're worth having there though, because even if they don't work, they don't convey any less information than they weren't links in the first place. Hopefully, in the not-too-distant future `rustdoc` will have a better and more robust solution for linking in crate docs.

AtheMathmo:
This all looks good to me! Will merge now.

Thanks for your hard work on this - I'm excited to make use of it!

Issue #54, opened 2016-09-26
Use new matrix! macro in unit tests


AtheMathmo:
With #47 merged we now have a macro for small matrices! This macro helps to enforce at compile time that our dimensions are correct for our input data and is stricter than the existing `new` constructor. Because of this we should use this for future tests and also convert existing tests to use this.

AtheMathmo:
Thanks @gcollura !

PR #73, opened 2016-10-24
Use macro matrix! in tests. Solve issue #54.

gcollura:
I hope it looks good.
I wanted to read the implementation and in the meantime I solved issue #54 .
If anything is missing, please tell me, I'll fix it.


tests/mat/mod.rs
```diff
@@ -7,7 +7,7 @@ fn test_l_triangular_solve_errs() {
     let a = Matrix::<f64>::new(0, 0, vec![]);
```
AtheMathmo: I think we can change this one too?

tests/mat/mod.rs
```diff
@@ -16,13 +16,15 @@ fn test_u_triangular_solve_errs() {
...
```
AtheMathmo: And here (+ probably a few other empty matrix cases?)

gcollura:
There's a bug with empty matrices, it cannot infer the type of the matrix (i32, f64, etc). This is the error:
```
$ cargo test
   Compiling rulinalg v0.3.4 (file:///xxxx/rulinalg)
error[E0282]: unable to infer enough type information about `_`
 --> <rulinalg macros>:2:37
  |
2 | { use $ crate :: matrix :: Matrix ; Matrix :: new ( 0 , 0 , vec ! [  ] ) } } ;
  |                                     ^^^^^^^^^^^^^ cannot infer type for `_`
tests/mat/mod.rs:17:13: 17:22 note: in this expansion of matrix! (defined in <rulinalg macros>)
  |
  = note: type annotations or generic parameter binding required

error: aborting due to previous error
```
This is why I left them as they were. Ideas?

Andlon:
@gcollura,

If you write

```rust
let mat = matrix![];
```

then - depending on what follows the above code - Rust may not be able to know which type the Matrix has (after all, you have not told it!), so you may need to be explicit:

```rust
let mat: Matrix<f64> = matrix![];
```

Note that this may not always be necessary, in which case we would prefer to avoid the explicit type specification.

gcollura:
Sorry for the missing ones.
I would leave some of the docs examples as they are and I would also leave such type of matrix declaration: `let a = Matrix::new(2, 3, vec![1.0; 6]); `, which are more compact, but still easily readable.
What do you think?

AtheMathmo:
I think that makes sense! Thanks

Andlon:
@AtheMathmo:

this kind of matrix construction seems to come up a lot, at least in our tests. Perhaps we should introduce a factory method like `Matrix::constant(rows, cols, value)`...?

AtheMathmo:
@Andlon thay may be worthwhile, though the alternative is:

```rust
const * Matrix::ones(rows, cols)
```

+/- some explicit typing

Andlon:
@AtheMathmo,
ah, I forgot about that. That's probably just as good. It's certainly better than the `let a = Matrix::new(rows, cols, vec![value; rows * cols]);` approach, since it's a lot more obvious!

gcollura:
Is it by design the scalar multiplication only compiles when the scalar comes after the matrix such as: `let mat: Matrix<f64> = Matrix::ones(rows, cols) * 3.` and not `let mat: Matrix<f64> = 3. * Matrix::ones(rows, cols)`? I may address this issue in another PR.

Andlon:
@gcollura,

Good catch, this seems to be an omission. At least, I couldn't find any implementation of `Mul<Matrix<T>> for f64` when searching the sources (I might have missed something though).

@AtheMathmo: Perhaps create a new issue so make sure we have implementations of all binary operations for both directions?

AtheMathmo:
I believe there was a reason I didn't do this, maybe generics or some other restriction made it impossible? I'd be very happy if someone else could figure it out though.

Andlon:
@AtheMathmo,

interesting. That might very well be! I keep getting tripped up by Rust's (implicit?) cross-crate rules. I'd like to look into this, but I think I better try to finish #71 first. Between thesis work and travelling I've been pretty busy lately. If anyone else wants to look into it, that would also be great!

gcollura:
Now it should be good to go. If I still missed something, please tell me!
Thanks

AtheMathmo:
It looks good at a glance @gcollura - thanks! I'll try to give it a thorough read later today.

@Andlon yes those get me too. I'm not sure if they interact here or not, but they might cause problems as `Mul` is defined out of crate and can be implemented on 'f64' (for example) in other crates. Of course 'Matrix' is internal, but I'm not sure how rustc handles that.

And no need to worry about any delays. In case you haven't noticed I'm also lagging behind, organizing visas and many other things for the start of my own PhD :).

Andlon:
@gcollura, looks good to me! Appreciate that you took the effort to do this!

@AtheMathmo, sounds exciting! Hope all goes well with the preparations and whatnot :)

src/matrix/decomposition/eigen.rs
```diff
@@ -285,7 +285,8 @@ impl<T: Any + Float + Signed> Matrix<T> {
             }
 
             let (c, s) = Matrix::givens_rot(x, y);
-            let givens_mat = Matrix::new(2, 2, vec![c, -s, s, c]);
+            let givens_mat = matrix!(c, -s;
```
AtheMathmo: This is the only, incredibly minor, thing that I noticed. I don't mind this change in itself, but we keep the `Matrix::new` construction on the givens matrices in the `svd` module. I think I prefer slightly here to keep it as the `Matrix::new`. Either way my opinion here isn't particularly strong and I'll happily merge as is.

PR #77, opened 2016-10-31
Sparse Matrix implementation

c410-f3r:
Hi again,

I am creating another PR due to the large amount of changes.
After a month of much struggle with a range of different things, I was able to create and fix the missing things for a basic and initial merge. Please, tell me if something is wrong/missing.

New things:
- Arithmetic
- Sparse row iterator
- CSC and CSR delegates to Compressed

Now I will improve the current code and start integrating the sparse matrix module with the BaseMatrix trait.

Thanks

### DIFF
--- a/benches/linalg/iter.rs
+++ b/benches/linalg/iter.rs
@@ -14,7 +14,7 @@ fn mat_diag_iter_10_50(b: &mut Bencher) {
     let a = Matrix::new(10, 50, vec![2.0;500]);
 
     b.iter(|| {
-    	let _ = black_box(a.iter_diag(DiagOffset::Main).cloned().collect::<Vec<_>>());
+        let _ = black_box(a.diag_iter(DiagOffset::Main).cloned().collect::<Vec<_>>());
     });
 }
 
@@ -24,10 +24,10 @@ fn mat_diag_manual_10_50(b: &mut Bencher) {
     let a = Matrix::new(10, 50, vec![2.0;500]);
 
     b.iter(|| {
-        let mut d = black_box( Vec::with_capacity(10));
+        let mut d = black_box(Vec::with_capacity(10));
         for i in 0..10 {
             unsafe {
-                black_box(d.push(*a.get_unchecked([i,i])));
+                black_box(d.push(*a.get_unchecked([i, i])));
             }
         }
     });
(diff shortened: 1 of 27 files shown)


AtheMathmo:
Really excited to check this out, thank you for your persistence with it!

AtheMathmo:
A brief look through makes me think this might be a little too big to merge all at once. As it is all new code it could be ok - but I'd still like to be able to ensure correctness absolutely. If it is too large I'll try to advise a sensible way to split. I just wanted to give you a heads-up so you can think about it in advance a little.

Andlon:
@c410-f3r: wow, this is big. I just skimmed very quickly over, and I really like what I see! Great job! I'm pretty busy these days, so I'm not sure when I'll have time to do a thorough review, but I'll try to look into it soon (and also participate in the discussion here, of course).

As for integrating with the BaseMatrix trait, I think perhaps it might be wise to wait a little, and it's something that probably needs more discussion. I believe the `BaseMatrix` trait would need a major redesign in order to accommodate this. From my understanding, `BaseMatrix` is primarily a device to reduce code duplication between `Matrix`, `MatrixSlice` and `MatrixSliceMut` (correct me if I'm wrong, @AtheMathmo ?), and I suppose sparse matrices don't fit so well into that particular purpose.

@AtheMathmo: "ensure correctness absolutely" is somewhat ambitious, since I don't think that can be said about any of the linear algebra implementations at the moment :P I suppose you mean you want to divide it into smaller chunks to make the actual _review_ easier? What potential problems do you anticipate with a large merge?

AtheMathmo:
> I believe the `BaseMatrix` trait would need a major redesign in order to accommodate this

Yes it probably would - or at least it may be difficult to meet some of the requirements for sparse matrices. Either way I agree that we should hold off on this for now. Historically these traits exist to make up for the fact that we can't do natural dereferencing because we need custom dynamically sized types. But the traits do cut down on code repetition primarily.

> "ensure correctness absolutely" is somewhat ambitious, since I don't think that can be said about any of the linear algebra implementations at the moment :P

Point well made!

I did mean to divide it up to make reviewing easier. Of course bugs are inevitable (and frankly more present in my own code than most other contributors) - but it is definitely easier to spot them if we review piece by piece.

Andlon:
@AtheMathmo, totally agree!

c410-f3r:
Closing this PR to split code is smaller patches

PR #78, opened 2016-11-01
Rename iterator functions to *_iter

c410-f3r:
Code has some automatic code formatting. Compilation and all tests are OK.

### DIFF
--- a/src/convert.rs
+++ b/src/convert.rs
@@ -16,7 +16,9 @@ impl<T> From<Vec<T>> for Vector<T> {
     }
 }
 
-impl<'a, T> From<&'a [T]> for Vector<T> where T: Clone {
+impl<'a, T> From<&'a [T]> for Vector<T>
+    where T: Clone
+{
     fn from(slice: &'a [T]) -> Self {
         Vector::new(slice.to_owned())
     }
@@ -32,7 +34,7 @@ macro_rules! impl_matrix_from {
     ($slice_type:ident) => {
         impl<'a, T: Copy> From<$slice_type<'a, T>> for Matrix<T> {
             fn from(slice: $slice_type<'a, T>) -> Self {
-                slice.iter_rows().collect::<Matrix<T>>()
+                slice.row_iter().collect::<Matrix<T>>()
             }
         }
     }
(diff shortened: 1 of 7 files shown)


AtheMathmo:
Thanks! Though this wont get merged for a while. As this contains breaking changes I'll be waiting until we have some other changes ready for a 0.4 release.

c410-f3r:
Can we rename `row_iter` to `rows_iter`?

AtheMathmo:
Personally I slightly prefer `row_iter`. Happy to hear from others - I'd like to stick with whichever most closely follows the rest of the community's naming patterns.

tafia:
I tend to prefer rows_iter ...
Even better now that we are doing a breaking change:
- rename rows in rows_len or height
- use rows for the iterator

AtheMathmo:
@tafia thanks for the comment.

>  rename rows in rows_len or height

You mean the `Matrix` field? If so I much prefer `rows`, but certainly wont shut it down if everyone disagrees.

> use rows for the iterator

I assume you mean the `iter()` function and `IntoIter` implementation? This was [briefly discussed](https://github.com/AtheMathmo/rusty-machine/issues/94) before - though again I'm happy to reopen the discussion. I'm come to be pretty happy with the choice to have the _default_ iterators be over all data. I think it is good to have an explicit row iterator function, and also a default `iter` is expected. We definitely want to be able to iterate over the data - as this is not trivial to do for `MatrixSlice` and I'm not sure what else we would call this function if not `iter`.

tafia:
Sorry I'm not sure I followed.

Here is what I thought:
- `iter` iterates every element of the matrix (per *default*, with row major order) ... We may have this one on Matrix only as there is no incentive over a `FlatMap<Rows>` when memory is not contiguous
- `rows` iterates over the rows (returns a `Rows: Iterator<Item=Row>` and `struct Row { inner: &[T] }`, on `BaseMatrix`
- `columns` over the columns ... Similarly with something more complicated for `Column` item ( probably a `MatrixSlice`) on `BaseMatrix`
- no more `..._iter` functions at all as we always return an iterator or a `MatrixSlice`

I must confess that I haven't followed all the discussions so I may be totally off (whatever you choose is probably fine anyway)

Issue #74, opened 2016-10-25
Erroneous singular matrix in .solve?


bright-star:
I'm verifying the results of the FD port I have written using rulinalg, which consists of the following two files:
- https://gitlab.com/bright-star/electro/blob/master/examples/empty_rect.rs
- https://gitlab.com/bright-star/electro/blob/master/src/lib.rs

Running that example `cargo run --example empty_rect` with nightly gives an LUP-related error:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: 
Error { kind: DivByZero, error: StringError("Singular matrix found in LUP decomposition. A value in the diagonal of U == 0.0.") }', ../src/libcore/result.rs:799
```

It also prints out the matrix and vector it was trying to solve with, which I verified against the MATLAB source I am porting with the same params: https://github.com/bright-star/computational-em/blob/master/finite-difference/e1_uniform.m So I know that this a solvable system.

For my sanity's sake: these are identical matrices, right?

```
-4   1   0   1   0   0   0   0   0
 1  -4   1   0   1   0   0   0   0
 0   1  -4   0   0   1   0   0   0
 1   0   0  -4   1   0   1   0   0
 0   1   0   1  -4   1   0   1   0
 0   0   1   0   1  -4   0   0   1
 0   0   0   1   0   0  -4   1   0
 0   0   0   0   1   0   1  -4   1
 0   0   0   0   0   1   0   1  -4

-4   1   0   1   0   0   0   0   0
 1  -4   1   0   1   0   0   0   0
 0   1  -4   0   0   1   0   0   0
 1   0   0  -4   1   0   1   0   0
 0   1   0   1  -4   1   0   1   0
 0   0   1   0   1  -4   0   0   1
 0   0   0   1   0   0  -4   1   0
 0   0   0   0   1   0   1  -4   1
 0   0   0   0   0   1   0   1  -4
```

AtheMathmo:
These are indeed the same matrix...

Could you give the vector too? It's possible this gives an opportunity to improve the stability of either LUP or the solver.

bright-star:
The vector is:

```
-100
0
0
-100
0
0
-100
0
0
```

AtheMathmo:
I spent a little time looking into this. I think there is a bug in the current implementation - the pivoting looks wrong to me (but the algorithm is working in most cases...).

I'll try to figure it out, I just wanted to let you know I'd taken a look.

bright-star:
[...]

PR #80, opened 2016-11-01
Fixing LU pivot bugs

AtheMathmo:
This resolves issue #74 (pending approval from @bright-star ).

There was an issue with the previous pivoting - often times the pivot matrix would not actually pivot correctly (while still producing a valid orthogonal matrix and so not failing in every case). The new pivot technique works directly on the columns which may be slightly less efficient (we also remove a transpose) - but is _more_ correct.

### DIFF
--- a/src/matrix/decomposition/lu.rs
+++ b/src/matrix/decomposition/lu.rs
@@ -1,13 +1,39 @@
-use matrix::{Matrix, BaseMatrix};
+use matrix::{Matrix, BaseMatrix, BaseMatrixMut};
 use error::{Error, ErrorKind};
 use utils;
 
 use std::any::Any;
 
-use libnum::{Float};
+use libnum::Float;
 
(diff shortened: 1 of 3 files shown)


Andlon:
Could you perhaps also include a test of an example that failed with the old implementation, so that it we can prevent future regressions?

AtheMathmo:
Thanks for reminding me!

I added a regression test for `solve` - which is what reported the error in #74. I also added a sanity test for the LUP decomp of the matrix in question, just to check it doesn't error (as it did before).

bright-star:
I'll try this ASAP, thanks!

bright-star:
No more errors for me! I haven't numerically verified it yet, though.

AtheMathmo:
Great! I numerically verified the example you gave in the issue. Let me know if you want to additional checks - if so I'll merge afterwards.

Issue #76, opened 2016-10-30
The `BaseMatrix` `diag` function should return an iterator instead of a Vec


AtheMathmo:
This proposed change gives an easier access point to more efficient actions on the main diagonal. This function will be equivalent to the new `mat.iter_diag(DiagOffset::Main)` function.

The current `diag` function copies the diagonal into a new `Vec`. This functionality can be easily retained using: `mat.diag().collect::<Vec<_>>()`.

AtheMathmo:
Closed with #82

PR #82, opened 2016-11-01
The `BaseMatrix` `diag` function should return an iterator (#76)

c410-f3r:
In `diag` method documentation I wasn't able to `assert_eq!` raw values since the iterator returns references of T. Is there another way?

### DIFF
--- a/src/matrix/slice.rs
+++ b/src/matrix/slice.rs
@@ -668,21 +668,23 @@ pub trait BaseMatrix<T>: Sized {
     /// Examples
     ///
     /// ```
-    /// #[macro_use]
-    /// extern crate rulinalg;
+    /// # #[macro_use]
+    /// # extern crate rulinalg;
     ///
     /// use rulinalg::vector::Vector;
     /// use rulinalg::matrix::{Matrix, BaseMatrix};
     ///
-    /// fn main() {
-    ///     let a = matrix!(1,2,3; 4,5,6; 7,8,9).diag().cloned().collect::<Vec<_>>();
-    ///     let b = matrix!(1,2; 3,4; 5,6).diag().cloned().collect::<Vec<_>>();
-    ///     let c = matrix!(1,2,3; 4,5,6).diag().cloned().collect::<Vec<_>>();
+    /// # fn main() {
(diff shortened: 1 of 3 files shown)


AtheMathmo:
You need to call `.cloned()` to clone the reference values.

Thanks for this! Will review soon but wont merge until other 0.4 changes are ready, as with #78 .

src/matrix/slice.rs
```diff
@@ -597,18 +597,17 @@ pub trait BaseMatrix<T>: Sized {
     /// let b = Matrix::new(3,2,vec![1,2,3,4,5,6]);
     /// let c = Matrix::new(2,3,vec![1,2,3,4,5,6]);
     ///
-    /// let d = &a.diag(); // 1,5,9
-    /// let e = &b.diag(); // 1,4
-    /// let f = &c.diag(); // 1,5
+    /// let d = a.diag().cloned().collect::<Vec<_>>(); // 1,5,9
```
AtheMathmo: I think this example was a little bit dodgy before you got around to it. I would suggest that we modify the example to collect the diagonal of a single matrix once? Let's also use the `matrix!` macro so that it is a little easier to see the diagonal.

AtheMathmo: @c410-f3r this is the only thing I am waiting on before merging. Everything else is good.

Do you have some time to tackle this soon? If not I am happy to complete it myself.

c410-f3r:
Ahhhhhh!!! I swear that I will never ever again say that I will publish any code in X days. Starting to feeling ashamed of myself >.>

AtheMathmo:
Don't worry about it - I do the same myself regularly :D

AtheMathmo:
This looks good - but there is one _so very tiny_ thing that I need to check.

It looks like the main body of the example is indented moreso than the rest. I want to compile the docs and make sure that it looks ok before merging.

AtheMathmo:
I finally figured out how to push my own minor tweaks to contributor PRs (when they have it enabled).

I hope you don't mind, I fixed the indenting as it was so minor. Will merge this now!

PR #83, opened 2016-11-01
Sparse Matrix implementation

c410-f3r:
Same PR, but with no arithmetic operations.

### DIFF
--- a/src/sparse_matrix/compressed_matrix/compressed/compressed_utils.rs
+++ b/src/sparse_matrix/compressed_matrix/compressed/compressed_utils.rs
@@ -1,4 +1,22 @@
-pub fn expand_ptrs<C>(ptrs: &[usize], mut closure: C)
+/// Given a slice of pointers, returns a closure that indicates its overlapped "index" and current
+/// expanded "value".
+/// If, for example, this function was used with the pointers of a `CSCMatrix`, the "index" would
+/// represent a row index and the "value" a index used to locate data and column index.
+///
+/// # Ilustrations
+///
+/// 0 , 3 , 6 , 8 , 8
+/// |___|___|___|___|
+///   |   |   |   |
+///   0   1   2   3
+///   |   |   |
+///   |   |   |-> (2, 6) - (2, 7)
+///   |   |-> (1, 3) - (1, 4) - (1, 5)
+///   |-> (0, 0) - (0, 1) - (0, 2)
+///
+/// All values are reverted, which gives us the following sequence: (2, 7) - (2, 6) - (1, 5) -
+/// (1, 4) - (1, 3) - (0, 2) - (0, 1) - (0, 0).
+pub fn expand_ptrs_rev<C>(ptrs: &[usize], mut closure: C)
     where C: FnMut(usize, usize)
 {
     for (idx, range) in ptrs.windows(2).enumerate().rev() {
--- a/src/sparse_matrix/compressed_matrix/compressed/compressed_utils.rs
+++ b/src/sparse_matrix/compressed_matrix/compressed/compressed_utils.rs
@@ -1,0 +1,9 @@
+pub fn expand_ptrs<C>(ptrs: &[usize], mut closure: C)
+    where C: FnMut(usize, usize)
+{
+    for (idx, range) in ptrs.windows(2).enumerate().rev() {
+        for nnz_idx in (range[0]..range[1]).rev() {
+            closure(idx, nnz_idx)
+        }
+    }
+}
--- a/src/sparse_matrix/compressed_matrix/compressed/mod.rs
+++ b/src/sparse_matrix/compressed_matrix/compressed/mod.rs
@@ -35,10 +35,12 @@ impl<T: Copy + One + Zero> CompressedMatrix<T> for Compressed<T> {
         let mut ptrs = vec![0; rows + 1];
         let mut sum_ptrs = 0;
 
+        // Fill ptrs with each row nnz
         for idcs_idx in 0..nnz {
             ptrs[triplets[idcs_idx].row()] += 1;
         }
 
+        // Incremental addition of rows nnz
         for ptr_idx in 0..rows {
             let tmp_ptr = ptrs[ptr_idx];
             ptrs[ptr_idx] = sum_ptrs;
@@ -47,6 +49,7 @@ impl<T: Copy + One + Zero> CompressedMatrix<T> for Compressed<T> {
 
         ptrs[rows] = nnz;
 
+        // Fill indices, data and adjust ptrs
         for idcs_idx in 0..nnz {
             let ptr_idx = triplets[idcs_idx].row();
             let dest_idx = ptrs[ptr_idx];
@@ -57,6 +60,7 @@ impl<T: Copy + One + Zero> CompressedMatrix<T> for Compressed<T> {
             ptrs[ptr_idx] += 1;
         }
 
+        // Correct ptrs order
         for ptr_idx in 0..nnz {
             let tmp_ptr = ptrs[ptr_idx];
             ptrs[ptr_idx] = last_ptr;
--- a/src/sparse_matrix/compressed_matrix/compressed/mod.rs
+++ b/src/sparse_matrix/compressed_matrix/compressed/mod.rs
@@ -7,8 +7,8 @@
 //! CSR matrix: Delegates all methods with no modifications.
 //! CSC matrix: Delegates all methods swaping rows for columns.
 
+mod compressed_iter;
 mod compressed_utils;
-mod compressed_iter_linear;
 
 use std::marker::PhantomData;
 use std::mem;
@@ -16,7 +16,7 @@ use std::mem;
 use libnum::{One, Zero};
 
 use self::compressed_utils::*;
-use sparse_matrix::{CompressedLinear, CompressedLinearMut, CompressedMatrix, Triplet, SparseMatrix};
+use sparse_matrix::{CompressedIter, CompressedIterMut, CompressedMatrix, Triplet, SparseMatrix};
 use sparse_matrix::compressed_matrix::Compressed;
 
 impl<T: Copy + One + Zero> CompressedMatrix<T> for Compressed<T> {
@@ -81,8 +81,8 @@ impl<T: Copy + One + Zero> CompressedMatrix<T> for Compressed<T> {
         self.data
     }
 
-    fn iter_linear(&self) -> CompressedLinear<T> {
-        CompressedLinear {
+    fn iter(&self) -> CompressedIter<T> {
+        CompressedIter {
             _marker: PhantomData::<&T>,
             current_pos: 0,
             data: self.data.as_ptr(),
@@ -92,8 +92,8 @@ impl<T: Copy + One + Zero> CompressedMatrix<T> for Compressed<T> {
         }
     }
 
-    fn iter_linear_mut(&mut self) -> CompressedLinearMut<T> {
-        CompressedLinearMut {
+    fn iter_mut(&mut self) -> CompressedIterMut<T> {
+        CompressedIterMut {
             _marker: PhantomData::<&mut T>,
             current_pos: 0,
             data: self.data.as_mut_ptr(),
--- a/src/sparse_matrix/compressed_matrix/compressed/mod.rs
+++ b/src/sparse_matrix/compressed_matrix/compressed/mod.rs
@@ -1,0 +1,240 @@
+//! The Compressed sparse matrix module (Row-major order).
+//!
+//! Since CSC and CSR matrices shares much of the same source code, this module is intended
+//! as a central destination for logical operations.
+//!
+//! How the delegation of responsibility works:
+//! CSR matrix: Delegates all methods with no modifications.
+//! CSC matrix: Delegates all methods swaping rows for columns.
+
+mod compressed_utils;
+mod compressed_iter_linear;
+
+use std::marker::PhantomData;
+use std::mem;
+
+use libnum::{One, Zero};
+
+use self::compressed_utils::*;
+use sparse_matrix::{Triplet, CompressedMatrix, SparseMatrix};
+
+/// The `Compressed` struct.
+///
+/// Can be instantiated with any type.
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub struct Compressed<T> {
+    cols: usize,
+    data: Vec<T>,
+    indices: Vec<usize>,
+    ptrs: Vec<usize>,
+    rows: usize,
+}
+
+/// Compressed matrix linear iterator
+#[derive(Debug)]
+pub struct CompressedLinear<'a, T: 'a> {
+    _marker: PhantomData<&'a T>,
+    current_pos: usize,
+    data: *const T,
+    indices: &'a [usize],
+    positions: usize,
+    ptrs: &'a [usize],
+}
+
+/// Compressed matrix mutable linear iterator
+#[derive(Debug)]
+pub struct CompressedLinearMut<'a, T: 'a> {
+    _marker: PhantomData<&'a mut T>,
+    current_pos: usize,
+    data: *mut T,
+    indices: &'a [usize],
+    positions: usize,
+    ptrs: &'a [usize],
+}
+
+impl<T: Copy + One + Zero> CompressedMatrix<T> for Compressed<T> {
+    fn from_triplets<R>(rows: usize, cols: usize, triplets: &[R]) -> Compressed<T>
+        where R: Triplet<T>
+    {
+        let nnz = triplets.len();
+
+        let mut data = vec![T::zero(); nnz];
+        let mut indices = vec![0; nnz];
+        let mut last_ptr = 0;
+        let mut ptrs = vec![0; rows + 1];
+        let mut sum_ptrs = 0;
+
+        for idcs_idx in 0..nnz {
+            ptrs[triplets[idcs_idx].row()] += 1;
+        }
+
+        for ptr_idx in 0..rows {
+            let tmp_ptr = ptrs[ptr_idx];
+            ptrs[ptr_idx] = sum_ptrs;
+            sum_ptrs += tmp_ptr;
+        }
+
+        ptrs[rows] = nnz;
+
+        for idcs_idx in 0..nnz {
+            let ptr_idx = triplets[idcs_idx].row();
+            let dest_idx = ptrs[ptr_idx];
+
+            data[dest_idx] = triplets[idcs_idx].value();
+            indices[dest_idx] = triplets[idcs_idx].col();
+
+            ptrs[ptr_idx] += 1;
+        }
+
+        for ptr_idx in 0..nnz {
+            let tmp_ptr = ptrs[ptr_idx];
+            ptrs[ptr_idx] = last_ptr;
+            last_ptr = tmp_ptr;
+        }
+
+        Compressed {
+            cols: cols,
+            data: data,
+            indices: indices,
+            ptrs: ptrs,
+            rows: rows,
+        }
+    }
+    fn new(rows: usize,
+           cols: usize,
+           data: Vec<T>,
+           indices: Vec<usize>,
+           ptrs: Vec<usize>)
+           -> Compressed<T> {
+        assert!(ptrs.len() == rows + 1, "Invalid pointers length");
+        assert!(data.len() == indices.len(),
+                "Data length must be equal indices length");
+
+        Compressed {
+            cols: cols,
+            data: data,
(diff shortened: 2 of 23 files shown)


AtheMathmo:
Great! Thanks for splitting these out. Will try to review this later today.

src/sparse_matrix/compressed_matrix/csc_matrix/mod.rs
```diff
@@ -0,0 +1,171 @@
...
+        self.compressed.nnz()
+    }
+    fn rows(&self) -> usize {
+        self.compressed.cols()
+    }
+
+    fn get(&self, row: usize, col: usize) -> T {
+        self.compressed.get(col, row)
+    }
+    fn transpose(&mut self) {
```
AtheMathmo: Just picking this as an example - in a few places we're missing new lines between functions. Obviously it's very minor but I'd prefer if you could add lines with any further changes.

src/sparse_matrix/compressed_matrix/mod.rs
```diff
@@ -0,0 +1,84 @@
...
+    fn from_triplets<R>(rows: usize, cols: usize, triplets: &[R]) -> Self where R: Triplet<T>;
+    /// Construct a new matrix based only in rows and cols lengh
+    fn new(rows: usize, cols: usize, data: Vec<T>, indices: Vec<usize>, ptrs: Vec<usize>) -> Self;
+
+    /// Returns indices
+    fn indices(&self) -> &[usize];
+    /// Returns pointers (offsets)
+    fn ptrs(&self) -> &[usize];
+
+    /// Iterates linearly over the the matrix returning its data and respective indices.
```
AtheMathmo: Another minor point, you typed 'the' twice here.

src/sparse_matrix/compressed_matrix/mod.rs
```diff
@@ -0,0 +1,84 @@
...
+    /// let a = CsrMatrix::new(3,3, vec![1, 2, 3], vec![0, 0, 1], vec![0, 2, 3, 3]);
+    ///
+    /// // Iterates linearly over CSR matrix, i.e, over rows
+    /// // Prints [0,0], [1,2] - [1], [3], [], []
+    /// for (row_cols_indices, row_data) in a.iter_linear() {
+    ///     println!("{:?}, {:?}", row_cols_indices, row_data);
+    /// }
+    /// ```
+    fn iter_linear(&self) -> CompressedLinear<T>;
+    /// Iterates linearly over of the matrix returning its mutable data and respective column indices.
```
AtheMathmo: Another typo here: 'linearly over the matrix' ?

src/sparse_matrix/triplet.rs
```diff
@@ -0,0 +1,23 @@
...
+pub trait Triplet<T> {
+    /// Returns column coordinate
+    fn col(&self) -> usize;
+    /// Returns row coordinate
+    fn row(&self) -> usize;
+    /// Returns value
+    fn value(&self) -> T;
+}
+
+impl<T: Copy> Triplet<T> for (usize, usize, T) {
```
AtheMathmo: Maybe we should require `Clone` here?

Alternatively, could we have `value()` return an `&T`? Or just consume the `Triplet` on value?

All pretty minor questions. I can probably figure them out myself as I digest more of the code :).

AtheMathmo: Looking at the `from_triplets` implementation I think it is better to consume the triplet when we get its value. Right now we can only use `from_triplets` if the underlying data type can be copied but we can remove this restriction safely I think.

Also I think the `from_triplets` function may do better if it takes an `ExactSizeIterator` instead of a slice. This covers slices and more general structures. I think this should work out fine but I haven't tried to implement it myself.

src/sparse_matrix/compressed_matrix/compressed/compressed_utils.rs
```diff
@@ -0,0 +1,9 @@
+pub fn expand_ptrs<C>(ptrs: &[usize], mut closure: C)
+    where C: FnMut(usize, usize)
+{
+    for (idx, range) in ptrs.windows(2).enumerate().rev() {
+        for nnz_idx in (range[0]..range[1]).rev() {
+            closure(idx, nnz_idx)
+        }
```
Andlon: I don't really have any idea what this function does. Perhaps document it a bit?

c410-f3r:
Thanks @AtheMathmo  and @Andlon. I will soon modify the code.

src/sparse_matrix/mod.rs
```diff
@@ -0,0 +1,47 @@
...
+/// Contract for sparse matrices implementation
+pub trait SparseMatrix<T> {
+    /// Constructs matrix with given diagonal.
+    ///
+    /// Requires slice of diagonal elements.
+    fn from_diag(diag: &[T]) -> Self;
+    /// Constructs the identity matrix.
+    ///
+    /// Requires the size of the matrix.
+    fn identity(size: usize) -> Self;
```
Andlon: Just want to point out that this is a minor point, and not so important right now :)

AtheMathmo:
Just to add to some of @Andlon 's comments - I think some inline documentation would be really valuable. I managed to figure most things out but it added some time. If you can't maintain this code for whatever reason we need other people to be able to jump in and understand it quickly.

Andlon:
Also, would it make sense to make the "linear" iterator simply `sparse_matrix.iter()`, which would iterate over `(row, col, value)` triplets? This would generalize to any sparse matrix, and if we also implement `IntoIterator`, you could simply write

```rust
for (i, j, v) in mat {
    // Do something with each non-zero value v in position (i, j)
}
```

src/sparse_matrix/mod.rs
```diff
@@ -0,0 +1,47 @@
...
+    /// Returns number of rows
+    fn rows(&self) -> usize;
+
+    /// Returns the specified data
+    fn get(&self, row: usize, col: usize) -> T;
+    /// Tranposes the given matrix
+    fn transpose(&mut self);
+
+    /// Returns a non-mutable reference to the underlying data.
+    fn data(&self) -> &[T];
```
c410-f3r: Wow! Forgot about that. Nice catch.

src/sparse_matrix/compressed_matrix/mod.rs
```diff
@@ -0,0 +1,84 @@
...
+    /// # Examples
+    ///
+    /// ```
+    /// use rulinalg::sparse_matrix::{CompressedMatrix, CscMatrix};
+    ///
+    /// let a = CscMatrix::new(3,3, vec![1, 2, 3], vec![0, 0, 1], vec![0, 2, 3, 3]);
+    ///
+    /// // Iterates linearly over CSC matrix, i.e, over columns.
+    /// // Prints [0,0], [1,2] - [1], [3], [], []
+    /// for (col_rows_indices, col_data) in a.iter_linear() {
```
c410-f3r: Ideally the iterator should only return values, that would give to the user the freedom to choose what kind of indices he wants to have, thus, avoiding copying unnecessary data.

```
for values in csr_matrix.iter() {}

for (row_idx, values) in csr_matrix.iter().enumerate() {}

for (row_idx, (row_cols_indices, values)) in csr_matrix.iter().transversal_indices().enumerate() {}
```

Andlon: @c410-f3r: I think this is an important thing to discuss! Could you perhaps re-post your answer in the main thread? In-line review comments tend to get lost very easily.

src/sparse_matrix/compressed_matrix/compressed/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+        let mut result_data = T::zero();
+
+        for idx in self.ptrs[row]..self.ptrs[row + 1] {
+            if self.indices[idx] == col {
+                result_data = self.data[idx];
+                break;
+            }
+        }
+
+        return result_data;
```
c410-f3r: Binary search would be a big win but some operations might not preserve row/col ordination. It is something that I will analyze deeply.

src/sparse_matrix/compressed_matrix/mod.rs
c410-f3r: Sure!

c410-f3r:
Ideally the iterator should only return values, that would give to the user the freedom to choose what kind of indices he wants to have, thus, avoiding copying unnecessary data.

```
for values in csr_matrix.iter() {}

for (row_idx, values) in csr_matrix.iter().enumerate() {}

for (row_idx, (row_cols_indices, values)) in csr_matrix.iter().transversal_indices().enumerate() {}
```

Of course we could simply return (i, j, value) but I think it is topic that needs more discussion.

src/sparse_matrix/compressed_matrix/compressed/mod.rs
Andlon: I think we can enforce the order of indices. Do you have any operations in mind which might be problematic..?

Andlon:
@c410-f3r, I worry that this is not so user-friendly... What do you think about the following suggestion?

```rust
// Use IntoIterator to make the sparse matrix iterable
for (i, j, v) in mat {}

// Same as above, but explicitly call iter
for (i, j, v) in mat.iter() {}

// Iterate only over the non-zero values, with no indices
for v in mat.data_iter() {}

// Iterate over the rows. Note: this should only be present for CsrMatrix.
for (row_index, row) in csr_mat.row_iter() {
    // What is a row in this case? Specialized struct?

    // Iterate only over the non-zero values in the row
    for v in row.data_iter() {}
    
    // Not so sure about this one, but just to be consistent with the above
    for (col, v) in row {} 
}

// Analogously for CscMatrix and column iteration
```

With this pattern, you should be able to touch only the data you need, yet the (presumably most common) case of iterating over triplets is easily accessible.

Let me know what you think!

(Also, I think we don't really need the row iterators yet. Just pointing out that it's a possible solution to the problem you mentioned).

src/sparse_matrix/compressed_matrix/compressed/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+            last_ptr = tmp_ptr;
+        }
+
+        Compressed {
+            cols: cols,
+            data: data,
+            indices: indices,
+            ptrs: ptrs,
+            rows: rows,
+        }
```
c410-f3r: 1. All documentation are stored in traits, didn't put more documentation in structs since they aren't exposed in public docs. @AtheMathmo said something like that in my first PR.

2. Don't think so, https://is.gd/fo29D6.

src/sparse_matrix/mod.rs
c410-f3r: Nice, I will note this feature.

src/sparse_matrix/compressed_matrix/compressed/mod.rs
Andlon: @c410-f3r: sorry, I meant inline comments. I.e. not user-facing documentation, but standard code comments, such as

`// In rare cases, it is possibly that foo is not bar, so we must account for this`

Basically, comments for us developers to more easily understand what's going on. Ideally this should be apparent from the code itself (by making sure to split up functionality into appropriately named functions), but in some cases the code is too technical to split up further, and comments can help explain what's really going on here. I think that this is the case here.

To be clear, I don't mean to say that there's something wrong with your code, but the very nature of the issue is pretty technical, and unless one has prior experience with implementing these data formats, it can be pretty hard to get a feeling for what's going on!

Andlon: As for 2, I'm a little confused what you mean. Your playground example seems to only print a slice of triplets? Maybe I wasn't clear. I mean that if you pass a set of triplets which contain duplicate indices, this needs to be taken into account, and the result should be summed.

For example:

```rust
let triplets = vec![ (0, 0, 2), (0, 0, 3) ];
let mat = CsrMatrix::from_triplets(2, 2, &triplets);

assert_eq!(mat.get(0, 0), 5);
```

I think from reading the code that this is not the current behavior, is that right...?

c410-f3r: I misunderstood both of your suggestions, sorry.
I will put some comments in the code, as for the duplicated tuples, we could use `HashSet` instead of slices (https://is.gd/f8lehT) or scan the slice to create a new slice of unique tuples.

Andlon: Also, I just wanted to say that if this turns out to be very difficult, we could also go for an implementation that is not optimally efficient (but still exhibits correct behavior) for now and get it merged, and then create a new issue for a proper efficient implementation! I've never implemented these data structures in such detail before, so I don't know how difficult it is to get it right.

src/sparse_matrix/triplet.rs
c410-f3r: @AtheMathmo. I managed to do something like this: https://is.gd/KVjHxU. Is it worth it?

src/sparse_matrix/compressed_matrix/compressed/mod.rs
c410-f3r: No, you were pretty clear. I was so dumb that couldn't understand, sorry².
It is becoming pretty challenging to write just a simple algorithm to sum values repeated rows/cols, I will see this in the near future.

src/sparse_matrix/compressed_matrix/compressed/mod.rs
c410-f3r: I was working with a algorithm that was doing it but I don't remember which was right now. I must take a look into the code.

c410-f3r:
@Andlon, I am unable to think right now and I will work on your suggestion about the iterators in the near future.

src/sparse_matrix/compressed_matrix/compressed/mod.rs
c410-f3r: LOOOOOOL! NVM! Pointers must have a crescent order, it is in their own definition. Lol!

c410-f3r:
For the sake of organization, I am grouping the missing things that should be resolved before code merge:

- [ ]  **Make code easier to understand**. Sorry guys, but I will be a little reluctant about adding code comments, therefore, this topic needs more research of my part.

> Programmers are taught to comment their code: good code has lots of comments. Unfortunately, they are never taught why code needs comments: bad code requires lots of comments.
> - The Pragmatic programmer

> It takes only a few seconds of thought to explain most of your intent in code. In many
cases it’s simply a matter of creating a function that says the same thing as the comment
you want to write.
> - Clean Code

- [ ] **Binary search in `get` method**. If we can guarantee that all rows/cols indices of the array of indices are sorted in ascending order, independently of the operation (arithmetics, creations, solvers and functions), then, we can use a binary search in `get` method.

- [ ] **Refactor the iterators of the compressed matrix**. Considering that a _scattered_
 iterator will also be created, what should be the best returning type and how this type would be served.

- [ ] **Triplets**
    - [ ]  **Duplicate (i,j) entries in will be summed**.
    - [ ] **Except `ExactSizeIterator` in `from_triplets` method**.

AtheMathmo:
@c410-f3r thank you for the overview - it made it real easy to catch up on all this (very valuable) discussion.

> why code needs comments: bad code requires lots of comments.

I don't totally agree with this - in any case I wouldn't make the argument that your code is bad but I would say some comments would be helpful. I'm not suggesting you go through and comment every thing you're doing - that's not the case in rulinalg currently either. But as @Andlon pointed out before, this is very technical stuff that requires quite a bit of background knowledge. For example, if I want to understand your `transpose` function I have to figure out what all of the data structures correspond to. And if your language differs from say, Wikipedia, then it could be fairly challenging.

Of course the solution isn't to describe in detail what every data structure is and what we're doing every step of the way. But perhaps in transpose (again, just as an example) we could do something like:

```rust
// ...
ptrs[0] = 0;

// Some brief explanation of what this loop is doing
for idx in 0..self.rows {
    ptrs[idx + 1] = ptrs[idx] + counter[idx];
 }
// ...
```

**I don't want to make out like the code is in a sorry state. To me it looks very nearly ready to merge and this particular issue definitely isn't a killer.** I'm sure as I get more familiar they will be a non-issue. But right now I'm in the fortunate position of being able to understand how new contributors will feel when approaching this particular part of the project and I want to take advantage of that rare situation and try to improve (in ways that I've not been able to in my own code).

src/sparse_matrix/compressed_matrix/compressed/mod.rs
Andlon: Hey, don't beat yourself up about it! First, this _is_ a difficult task, and finding a correct and efficient algorithm here is not at all simple! Second, it's not obvious that summing up triplets is desirable unless you've worked with e.g. finite element analysis, where this is very common!

One stop-gap solution might be to sort the triplets, sum together duplicates (which are adjacent after the sort), and then use your existing implementation. This is not as efficient as it can get, but like I said, if we can get a working algorithm first, we can worry about efficiency later.

Andlon:
@c410-f3r: thanks for the overview! Definitely makes things a little easier to follow.

About the code comments part: I definitely agree with the idea, and the two books you quote are classics in the field. Generally, if you can, your code should be self-documenting. That is, the name of your functions and variables should explain what it is you're doing. Now, what it often cannot do, is explain _why_ you are doing something, or to justify a particular course of action. Moreover, you might find applying those principles difficult when writing low-level data structures (which includes the compressed matrix formats) and when working with mathematics, in which case things may be defined only mathematical concepts, and consequently coming up with good names might be excruciatingly difficult.

Regarding the _why_, consider the following silly example:

```rust
// First saddle the horse
{ ... }

// Put shoes on horse
{ ... }

// Give horse ice cream
{ ... }

// Ride horse
{ ... }
```

Now, using the principles you mentioned, we can greatly improve this without comments at all:

```rust
saddle_horse(&mut horse);
put_shoes_on_horse(&mut horse);
feed_horse_ice_cream(&mut horse);
person.ride(&horse);
```

Great! Our code is now self-documenting. But why the hell are we giving the horse ice cream before riding it? This calls for a comment!

```rust
saddle_horse(&mut horse);
put_shoes_on_horse(&mut horse);

// Our horse is unfortunately allergic to carrots, and is a very picky eater.
// From experience, the only thing that will make the horse cooperative is ice cream
feed_horse_ice_cream(&mut horse);
person.ride(&horse);
```

Ah, so that's why.

Now, I'm sure the code in question can be organized better into smaller functions. Just as an example, you might want to do something like `determine_sparsity_pattern()` in the `from_triplets` function, which first determines the indices and row pointers before putting the data in. I don't know if this particular example is appropriate, but this kind of refactoring might help make the code easier to read and understand!

src/sparse_matrix/triplet.rs
AtheMathmo: I am actually a little conflicted. Originally I thought we should remove the `Copy` bound on the `(usize, usize, T)` implementation. This works fine but means we cannot access `triplet.value()` within a loop as it is being borrowed from the `ExactIterator` object. In fact this is also the case when just the trait is changed: https://is.gd/6SFtvM .

I'm not sure how best to proceed. Perhaps we have `value(&self) -> &T` and we allow `from_triplets` to work on `T: Clone`. This way we cover all the primitive `Copy` types and can have more complex data structures if our users ever want it?

Andlon: @AtheMathmo: at a minimum, `T` would need `Clone` and `Add`. I think it can be done without `Copy`, yeah. However, it may perhaps be easier to just make a working implementation with `Copy` first? Then we can see if we can make the trait bounds less restrictive afterwards. It's actually a relatively difficult problem to implement.

c410-f3r:
Thanks @AtheMathmo and @Andlon.

AtheMathmo:
Hey @c410-f3r ! 

Just checking your status on this. No rush of course - but if you feel blocked is there anything I can help with? Alternatively did you want someone else to jump in and help out with this PR?

It's totally fine for you to take as long as you need with this :)

c410-f3r:
Everything is fine now that I completed my exams and tomorrow I will probably finish the missing features.

AtheMathmo:
I hope they went well!

I look forward to the updated PR.

bright-star:
Is this PR still ongoing? I'm very excited about it.

c410-f3r:
It is in a slow pace since I am busy with work, study and other stuff :/

AtheMathmo:
@bright-star it is a really cool PR!

@c410-f3r if you need any help completing this please don't hesitate to ask. You've put a lot of work in and I would be more than happy to help you make the remaining changes.

c410-f3r:
Thanks @AtheMathmo!

c410-f3r:
Lost track of the branch, again.

PR #84, opened 2016-11-02
Adding column iterators (#3)

c410-f3r:
For now, the iterator is returning `Vec<&T>` since it is not possible to return slices. Another possible returning type is `MatrixSlice` as defined in #3.
The column iterator will iterate the full defined slice shape because I didn't use `col_stride` as the row iterator uses `row_stride`.
There are some automatic formatting code.

### DIFF
--- a/src/matrix/iter.rs
+++ b/src/matrix/iter.rs
@@ -2,10 +2,9 @@ use std::iter::{ExactSizeIterator, FromIterator};
 use std::mem;
 use std::slice;
 
-use super::{Matrix, MatrixSlice, MatrixSliceMut, Rows, RowsMut, Diagonal, DiagonalMut};
+use super::{Cols, ColsMut, Diagonal, DiagonalMut, Matrix, MatrixSlice, MatrixSliceMut, Rows,
+            RowsMut};
 use super::slice::{BaseMatrix, BaseMatrixMut, SliceIter, SliceIterMut};
-
-
 
 macro_rules! impl_iter_diag (
     ($diag:ident, $diag_base:ident, $diag_type:ty, $as_ptr:ident) => (
@@ -37,7 +36,7 @@ impl<'a, T, M: $diag_base<T>> Iterator for $diag<'a, T, M> {
             None
         }
     }
-    
+
     fn nth(&mut self, n: usize) -> Option<Self::Item> {
         self.diag_pos += n * (self.matrix.row_stride() + 1);
         if self.diag_pos < self.diag_end {
@@ -67,7 +66,6 @@ impl<'a, T, M: $diag_base<T>> Iterator for $diag<'a, T, M> {
 }
 
 impl<'a, T, M: $diag_base<T>> ExactSizeIterator for $diag<'a, T, M> {}
-
     );
 
 );
@@ -143,6 +141,78 @@ impl_iter_rows!(RowsMut, &'a mut [T], from_raw_parts_mut);
 
 impl<'a, T> ExactSizeIterator for Rows<'a, T> {}
 impl<'a, T> ExactSizeIterator for RowsMut<'a, T> {}
+
+macro_rules! impl_iter_columns (
+    ($columns:ident, $data_type:ty, $get:ident) => (
+
+/// Iterates over the columns in the matrix.
+impl<'a, T> Iterator for $columns<'a, T> {
+    type Item = $data_type;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.col_pos >= self.slice_cols {
+            return None;
+        }
+
+        let mut column: $data_type = Vec::with_capacity(self.slice_rows);
+        unsafe {
+            for row_idx in 0..self.slice_rows {
+            	let ptr_idx = (self.col_pos + self.slice_rows * row_idx) as isize;
+                column.push(self.slice_start.offset(ptr_idx).$get().unwrap());
+            }
+        }
+        self.col_pos += 1;
+        Some(column)
+    }
+
+    fn last(self) -> Option<Self::Item> {
+        if self.col_pos >= self.slice_cols {
+            return None;
+        }
+
+		let last_col = self.slice_cols - 1;
+        let mut column: $data_type = Vec::with_capacity(self.slice_rows);
+        unsafe {
+            for row_idx in 0..self.slice_rows {
+            	let ptr_idx = (last_col + self.slice_rows * row_idx) as isize;
+                column.push(self.slice_start.offset(ptr_idx).$get().unwrap());
+            }
+        }
+        Some(column)
+    }
+
+    fn nth(&mut self, n: usize) -> Option<Self::Item> {
+        if self.col_pos + n >= self.slice_cols {
+            return None;
+        }
+
+        let mut column: $data_type = Vec::with_capacity(self.slice_rows);
+        unsafe {
+           for row_idx in 0..self.slice_rows {
+            	let ptr_idx = ((self.col_pos + n) + self.slice_rows  * row_idx) as isize;
+                column.push(self.slice_start.offset(ptr_idx).$get().unwrap());
+            }
+        }
+        self.col_pos += n + 1;
+        Some(column)
+    }
+
+    fn count(self) -> usize {
+        self.slice_rows - self.col_pos
+    }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        (self.slice_rows - self.col_pos, Some(self.slice_rows - self.col_pos))
+    }
+}
+    );
+);
+
+impl_iter_columns!(Cols, Vec<&'a T>, as_ref);
+impl_iter_columns!(ColsMut, Vec<&'a mut T>, as_mut);
+
+impl<'a, T> ExactSizeIterator for Cols<'a, T> {}
+impl<'a, T> ExactSizeIterator for ColsMut<'a, T> {}
 
 /// Creates a `Matrix` from an iterator over slices.
 ///
@@ -338,7 +408,7 @@ mod tests {
 
     #[test]
     fn test_empty_matrix_diag() {
-        let a : Matrix<f32> = matrix![];
+        let a: Matrix<f32> = matrix![];
 
         assert_eq!(None, a.iter_diag(DiagOffset::Main).next());
     }
@@ -401,7 +471,7 @@ mod tests {
         let a = matrix![0.0, 1.0, 2.0, 3.0;
                         4.0, 5.0, 6.0, 7.0;
                         8.0, 9.0, 10.0, 11.0];
-        let b = MatrixSlice::from_matrix(&a, [0,0], 2, 4);
+        let b = MatrixSlice::from_matrix(&a, [0, 0], 2, 4);
 
         let mut diags_iter = b.iter_diag(DiagOffset::Main);
         assert_eq!(5.0, *diags_iter.nth(1).unwrap());;
@@ -413,7 +483,7 @@ mod tests {
 
         let mut diags_iter = b.iter_diag(DiagOffset::Below(1));
         assert_eq!(4.0, *diags_iter.nth(0).unwrap());
-        assert_eq!(None, diags_iter.next());        
+        assert_eq!(None, diags_iter.next());
     }
 
     #[test]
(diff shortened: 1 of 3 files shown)


src/matrix/iter.rs
```diff
@@ -144,6 +142,78 @@ impl_iter_rows!(RowsMut, &'a mut [T], from_raw_parts_mut);
...
+        if self.col_pos >= self.slice_cols {
+            return None;
+        }
+
+		let last_col = self.slice_cols - 1;
+        let mut column: $data_type = Vec::with_capacity(self.slice_rows);
+        unsafe {
+            for row_idx in 0..self.slice_rows {
+            	let ptr_idx = (last_col + self.slice_rows * row_idx) as isize;
+                column.push(self.slice_start.offset(ptr_idx).$get().unwrap());
```
AtheMathmo: Though less safe - I think it is better to use `mem::transmute` here. Both `get_mut` and `get_ref` do a null check. As we're sure the pointer is valid, we should remove the null check.

Ideally we would use [proper dereferencing](https://doc.rust-lang.org/src/core/up/src/libcore/ptr.rs.html#325-331): `&*ptr` and `&mut *ptr`.

src/matrix/iter.rs
```diff
@@ -637,7 +707,107 @@ mod tests {
     }
 
     #[test]
-    fn into_iter_compile() { 
+    fn test_matrix_cols() {
```
AtheMathmo: We should also include some tests for `MatrixSlice` - just to make sure there will be no regressions there in the future.

src/matrix/slice.rs
```diff
@@ -168,6 +168,30 @@ pub trait BaseMatrix<T>: Sized {
         }
     }
 
+    /// Iterate over the columns of the matrix.
```
AtheMathmo: I'm not sure about this but I think it might be worth stating that this access pattern will be _very_ sub-optimal. We'll have lots of cache misses (on larger matrices) and as the data is non-contiguous will miss out on SIMD optimizations.

Andlon:
Actually, I think `Vec<&T>` is not ideal. This way you're forcing a potentially large allocation for every column if you only want to access each value once, and you cannot easily i.e. subtract columns from each other or similar.

@AtheMathmo: Why do you feel that `Vec` is better than `MatrixSlice` in this case...? I'm also not entirely convinced that `iter_rows()` should not return a `MatrixSlice` in the first place. I would hope (?) Rust should be able to remove the bounds checks when iterating over a 1-row `MatrixSlice`, so in most cases the performance difference should be negligible (this would need testing), and `MatrixSlice` provides access to much greater utility. I think the inconsistency in the API here is problematic.

There are cases when you actually _need_ access to a contiguous slice, in which case we could alternatively provide a method on `MatrixSlice` that returns `Option<&[T]>`?

AtheMathmo:
You may be right.

> Why do you feel that Vec is better than MatrixSlice in this case...?

Truthfully because of the inconsistency we'd incur with the `iter_rows()` function.

> Rust should be able to remove the bounds checks when iterating over a 1-row `MatrixSlice`

We should test this - but when I was toying with it before I was disappointed to see that it was not optimized as well as the slice. From memory the bound checks were ok but it was a pain to get things to vectorize correctly. It's pretty difficult to get rust to automatically use SIMD instructions without falling back to working on slices - which is basically what the `utils` module is for. All that said, I think it is worth looking into and seeing what the actual difference would be. I agree that having `MatrixSlice` in both cases would be preferable.

> There are cases when you actually need access to a contiguous slice, in which case we could alternatively provide a method on MatrixSlice that returns Option<&[T]>?

How exactly would this work? We have [`get_row`](https://athemathmo.github.io/rulinalg/doc/rulinalg/matrix/slice/trait.BaseMatrix.html#method.get_row) already - I suppose this would get the job done? We could also have `get_row` return a `MatrixSlice`, and add `get_col` to do the same, finally adding a new function, `get_row_slice`?

Andlon:
@AtheMathmo: `get_row` is similar, but I was actually thinking of a method such as `MatrixSlice::as_contiguous_data() -> Option<&T>` (and also a mutable version), which only returns `Some` when the slice truly happens to be contiguous.

I'm confused when you talk about SIMD. Vectorize... What, exactly? In this case we're only talking about the return type for an iterator, so I presume you're talking about vectorization in functions that rely on this iterator? Which ones? And in those cases, could you perhaps still use `as_contiguous_data()` to recover the same functionality?

Andlon:
@AtheMathmo: ah, I think maybe I understand what you mean. See below example.

```rust
// Current version
let mut answer = 0.0f64;
for row in mat.iter_rows() {
    for element in row {
         // Is this the case where you're worried about losing auto-vectorization?
         answer += 2.0 * element;
    }
}
```

AtheMathmo:
I think I would prefer to have a function which would allow the user to pick any row as contiguous data. The `as_contiguous_data` would only return a value if the `MatrixSlice` is only one row, or is a full matrix. The former case we can cover with `get_row_as_slice`. The latter case I think confuses the `as_contiguous_data` as the user can't really be sure (by which I mean the type system doesn't help them) if they are getting a single row, or the whole matrix.

And yes I meant auto-vectorization as in your example. Right now we can also use the `utils` functions to guarantee vectorization on the rows (as they are already slices). We can get this back by converting our `MatrixSlice` to `&[T]` if needed of course.

As one final point: I _don't think_ this is an issue, but it's possible we are using `iter_rows` to do some of the `MatrixSlice` arithmetic, i.e. zipping row iterators and vectorizing the operation per row. If `iter_rows` returns a `MatrixSlice` we would need to change these.

Andlon:
My opinion at this point is that the API of `rulinalg` should not be designed around the current shortcomings of the Rust compiler (unless they happen to be shortcomings that cannot be fixed in future releases), but instead provide ways to work around these shortcomings when it makes sense.

I think returning `MatrixSlice` for the row/column iterators vastly improves the user experience, and we can recover the current functionality with one of the functions you mentioned. However, doing `row.get_row(0)` is very counter-intuitive, which is why I brought up the `as_contiguous_data()` idea, as you'd be writing `row.as_contiguous_data()` instead. As you point out, what exactly you get here is somewhat context-sensitive, which may not be optimal. However, `get_row_as_slice` is also clunky, as `get` indicates a lookup-operation, and slice is ambiguous as it might mean `std::slice` or `MatrixSlice`. Perhaps `as_contiguous_row(&self) -> Option<&[T]>` then...?

Ideally we'd provide a `Row` struct which implements `MatrixSlice`, but as usual the combinatorial explosion of binary operators is an issue. Unfortunately `Deref<MatrixSlice>` does not allow binary operators on the receiver...

AtheMathmo:
I definitely agree with what you're saying. As you point out, ideally we'd have a `Row` struct which would remove all ambiguity - though it would basically be a wrapper for a `&[T]` and we'd need a `Col` struct which would have some fairly complex code to fill the functionality of `Row` (and would be less efficient).

My issue with the `as_contiguous_data` was more that this function would be exposed on `MatrixSlice` at least and using it outside of the `iter_row` or `get_row` cases would be non-obvious I think.\* I worry that `as_contiguous_row` would have the same problem. If our `MatrixSlice` isn't just a row what should the user expect? Should it always get the first row? Fail if there is more than one row? Return all of the data if the `MatrixSlice` is a full contiguous matrix? These aren't big questions but it does make the approach feel a little less ideal to me compared to hiding this as another `get_row_*` function - which definitely does have the drawbacks you described.

I think we should definitely look at using `MatrixSlice` as the return value of the row/column iterators, and also the `get_row`(/`get_col`) functions. If that seems to be working well, I think we are both in agreement that we still need some easy way to get a `&[T]` for the row. I guess we have time to figure out how that should look.

\* To be clear, I think `mat_slice.get_row(3).as_contiguous_data()` looks great, but `mat_slice.as_contiguous_data()` would be confusing.

Andlon:
I'm sorry, but I don't see why `get_row_as_slice()` is less confusing than for example `as_contiguous_row()`? Outside of the mentioned scenarios, they seem both equally out of place, in my mind, and so they seem to suffer from the same problem.

Also, `as_contiguous_data()` would be something most users would not need to worry about. Only when you need to squeeze out extra performance would you have to care, and in that case all you need to know is that _if_ the slice happens to be represented purely as a continuous chunk of data, then it will return this chunk, otherwise `None`. If you know that your slice is a row (which would be the case when iterating over rows, or extracting a single row with `get_row`), then you'd get the row as a contiguous slice. I proposed `as_contiguous_data()` as opposed to something row-specific specifically because it makes sense for any `MatrixSlice`, and could conceivably be used also in other scenarios, though I must admit that I've failed to come up with another practical usage. I completely understand your worries, however, as I suppose it can definitely be confusing nonetheless.

In any case, all these suggestions are suboptimal in view of a specialized type that delegates to `MatrixSlice`, but that seems currently out of the question due to the binary operator problem, which is unfortunate. At any rate, I agree with your conclusion!

AtheMathmo:
@tafia posted something related in #78 . It makes me wonder whether we can achieve the best of both, by adding new `Row` and `Col` structs that look like this:

```rust
pub struct Row<'a, T> {
   inner: &'a [T]
} 

impl<T> Row<T> {
    pub fn to_mat(self) -> MatrixSlice { // do some unsafe `from_raw` stuff }
    
    pub fn as_contiguous_data(&self) -> &'a [T] { self.inner }
}
```

And then our `BaseMatrix` trait would look something like this:

```rust
pub fn get_row(idx: usize) -> Row;

pub fn get_row_unchecked(idx: usize) -> Row;

pub fn row_iter(&self) -> /* Iterator which goes over Row */
```

This way we don't have the explosion of binary ops implementations. We do have some added complication as we'd require: `Row`, `Col`, `RowMut`, `ColMut`. But it is easy for our user to choose whether they want to treat the row as a matrix or as a contiguous slice. 

Also the `Col` struct would be a little more complicated - maybe it would have it's `inner` as a `MatrixSlice` for the column directly?

It might be worth pushing this discussion into one issue?

Andlon:
@AtheMathmo: that's an interesting approach, but unfortunately it makes the user jump through hoops a little if he wants to perform `MatrixSlice` operations on the rows, for example. One of the reasons I want to use `MatrixSlice` directly, is that you could do things like:

```rust
// Add 2.0 to all elements of row 5
mat.get_row_mut(5) += 2.0;

// Contrast this with:
mat.get_row_mut(5).to_mat() += 2.0;

// By the way, this would be much neater with just `row_mut(5)`, no?
// (is the get redundant?)
```

It still works, of course, but it's added noise.

And yes, we should move this into a new issue, I think! Seems to be a lot to discuss here.

AtheMathmo:
I opened #87 .

Personally I think the additional function call isn't too gross. If the user does want the contiguous slice they have to jump through more hoops otherwise (or understand how our `to_contiguous_data` function works - not too big a deal though).

> // By the way, this would be much neater with just `row_mut(5)`, no?
> // (is the get redundant?)

I was under the impression I should use `get` because the function returns an option. At the very least I think it makes sense when couple with `get_row_unchecked` (I think `row_unchecked` is a bit less self-explanatory).

c410-f3r:
Thanks for the feedback @AtheMathmo and @Andlon.

AtheMathmo:
@c410-f3r Sorry that the PR got a little derailed compared to the issue. But thanks again for opening - it's spurred a lot of great discussion!

Andlon:
@AtheMathmo: (I'll just briefly answer to the last points here, then we can instead continue in #87). Sorry, I forgot it returns an `Option`. You're right, it all makes sense as it is!

In terms of jumping through hoops, I think the contiguous slice option is for advanced usage, as you'd rarely need it in practice, and I think the API should be designed towards the common usage. There will be a trade-off between ergonomy for the common case and the advanced case, and I think the focus should be on making the most common use case as ergonomic as possible (while letting the advanced case remain somewhat reasonably ergonomic still).

c410-f3r:
Just passing by to say that i am bit busy studying because of my college and soon I will submit more changes

Andlon:
@c410-f3r: thanks for the update! Take as much time as you need, and best of luck with your studies :)

AtheMathmo:
Hey @c410-f3r - we have #99 ready to merge now. As a result this PR will need some changes. There's no need to rush as this wont constitute a breaking change. I just wanted to let you know.

c410-f3r:
Thanks @AtheMathmo. You did a nice job working on this feature.

AtheMathmo:
Thank you! :)

AtheMathmo:
Sorry that this got a little side tracked @c410-f3r . I hope that now that we confirmed the approach with #99 you can deliver this feature a little more painlessly. It is definitely a good one!

PR #90, opened 2016-11-11
Fixed additional lu pivot bugs

mabruckner:
I was having some issues with .solve; the lu decomposition failed for some simple matrices that had obvious solutions (I added the smallest one to the lu test.) It turns out that there were still lingering bugs in the pivot selection code. It didn't take into account that the optimal pivot for a given column can be changed by previous steps in the decomposition, which in some cases meant that the selected pivot was zero while other candidates were non-zero. I fixed the problem by computing the pivot matrix at the same time as the lower and upper matrices.

It seems to work, but you might want to check it just to be sure - I had to change the entire decomposition method.

### DIFF
--- a/src/matrix/decomposition/lu.rs
+++ b/src/matrix/decomposition/lu.rs
@@ -40,7 +40,7 @@ impl<T> Matrix<T> where T: Any + Float
             let mut curr_max_idx = index;
             let mut curr_max = u[[curr_max_idx, curr_max_idx]];
 
-            for i in curr_max_idx..n {
+            for i in (curr_max_idx+1)..n {
                 if u[[i, index]].abs() > curr_max.abs() {
                     curr_max = u[[i, index]];
                     curr_max_idx = i;
--- a/src/matrix/decomposition/lu.rs
+++ b/src/matrix/decomposition/lu.rs
@@ -5,44 +5,6 @@ use std::any::Any;
 
 use libnum::Float;
 
-fn lup_decomp<T>(size:usize, l:&mut Matrix<T>, u:&mut Matrix<T>, p:&mut Matrix<T>) -> Result<(), Error> where T: Any + Float {
-    let n = l.rows();
-    if size == 0 {
-        Ok(())
-    } else {
-        let mut curr_max_idx = n-size;
-        let mut curr_max = u[[curr_max_idx, curr_max_idx]];
-
-        for i in curr_max_idx..n {
-            if u[[i, n-size]].abs() > curr_max.abs() {
-                curr_max = u[[i, n-size]];
-                curr_max_idx = i;
-            }
-        }
-        if curr_max.abs() < T::epsilon() {
-            return Err(Error::new(ErrorKind::DivByZero,
-                "Singular matrix found in LUP decomposition. \
-                A value in the diagonal of U == 0.0."));
-        }
-
-        if curr_max_idx != n-size {
-            l.swap_rows(n-size, curr_max_idx);
-            u.swap_rows(n-size, curr_max_idx);
-            p.swap_rows(n-size, curr_max_idx);
-        }
-        l[[n-size, n-size]] = T::one();
-        for i in (n-size+1)..n {
-            let mult = u[[i, n-size]]/curr_max;
-            l[[i, n-size]] = mult;
-            u[[i, n-size]] = T::zero();
-            for j in (n-size+1)..n {
-                u[[i, j]] = u[[i,j]] - mult*u[[n-size, j]];
-            }
-        }
-        lup_decomp(size-1, l, u, p)
-    }
-}
-
 impl<T> Matrix<T> where T: Any + Float
 {
     /// Computes L, U, and P for LUP decomposition.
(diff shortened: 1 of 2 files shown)


src/matrix/decomposition/lu.rs
```diff
@@ -5,35 +5,46 @@ use std::any::Any;
...
+            let mult = u[[i, n-size]]/curr_max;
+            l[[i, n-size]] = mult;
+            u[[i, n-size]] = T::zero();
+            for j in (n-size+1)..n {
+                u[[i, j]] = u[[i,j]] - mult*u[[n-size, j]];
             }
         }
-
-        p
+        lup_decomp(size-1, l, u, p)
```
bright-star: The latest thread I could find was a stalled ticket for TCO: https://github.com/rust-lang/rfcs/issues/271

Andlon:
@mabruckner: just wanted to pop in and say thank you for taking the time for both finding the bug and providing a fix!

I'm not intimately familiar with the details of LU decomposition at the moment, so I cannot comment on whether or not the algorithm looks as it should, but the fact that it passes tests (and especially the latest one you've added) seems to suggest so! I just made a review comment on the use of recursion. As I stated there, I'm worried that the recursive implementation may overflow the stack for large matrices. Have you by any chance tried it with some relatively large matrices to see what happens?

Perhaps it could be rewritten in an iterative manner?

mabruckner:
Done, it was pretty easy to modify the code to be iterative.

AtheMathmo:
This looks neat! It also has the advantage of being more efficient. If all the tests are passing I'm not sure what more I can do to check the algorithm. Do you have a reference I could compare against? It would also be helpful to include this in the module docs if you have one.

I've only looked through briefly but will do a more thorough check soon and then hopefully merge!

src/matrix/decomposition/lu.rs
```diff
@@ -59,47 +32,42 @@ impl<T> Matrix<T> where T: Any + Float
...
 
-            for j in 0..i+1 {
-                let mut s1 = T::zero();
+        for index in 0..n {
+            let mut curr_max_idx = index;
+            let mut curr_max = u[[curr_max_idx, curr_max_idx]];
 
-                for k in 0..j {
-                    s1 = s1 + l[[j, k]] * u[[k, i]];
+            for i in curr_max_idx..n {
```
AtheMathmo: Very minor: We can start 1 larger here right? The first check will always return the false as they are the same value (I think).

mabruckner:
Sorry, I wasn't really working from a reference, just experience with similar code.

AtheMathmo:
No problem, this looks good to me. Later I'll check it over again quickly and hopefully merge it in.

src/matrix/decomposition/lu.rs
```diff
@@ -59,47 +32,42 @@ impl<T> Matrix<T> where T: Any + Float
     /// - Matrix cannot be LUP decomposed.
     pub fn lup_decomp(&self) -> Result<(Matrix<T>, Matrix<T>, Matrix<T>), Error> {
         let n = self.cols;
-        assert!(self.rows == n, "Matrix must be square for LUP decomposition.");
```
AtheMathmo: I think we want to keep this assertion? If your algorithm does work for the general case then we should still do the following before merging:

1. Edit the documentation, it no longer panics if matrices aren't square.
2. Add more tests for non-square matrices.

mabruckner: I missed that assertion, I'll add it back in

AtheMathmo:
Thanks for your work on this, will merge now!

src/matrix/decomposition/lu.rs
```diff
@@ -60,46 +33,42 @@ impl<T> Matrix<T> where T: Any + Float
...
         assert!(self.rows == n, "Matrix must be square for LUP decomposition.");
-
         let mut l = Matrix::<T>::zeros(n, n);
-        let mut u = Matrix::<T>::zeros(n, n);
-        let p = self.produce_pivot();
-        let a_2 = &p * self;
-
-        for i in 0..n {
-            l[[i, i]] = T::one();
+        let mut u = self.clone();
```
AtheMathmo: This is a bit of a no-go in rust. We should try hard to avoid using `self.clone` and taking `&self` as the input argument. In this case it would be better to have the function take ownership of `self`. This way we avoid the `clone`, we're more efficient than the old version and we also give the user the choice of whether they want to clone `self` or not.

Sadly this is a breaking change so this PR will have to join the queue of PRs waiting for a 0.4 version bump. Hopefully it won't take long... Maybe it would be worth merging this with the `clone`, and removing it as part of the 0.4 release later.

mabruckner: It makes sense for the decomposition to take ownership, as it is mostly an in-place algorithm. I think that it would be best to merge in the current version and make a separate PR for the 0.4 release that changes the signature and removes the `clone`.

Andlon: Agree that this should get merged now, and improved later. I think correctness needs to come before performance (within reason)!

PR #92, opened 2016-11-20
Fix ULP comparator bug

Andlon:
First `assert_matrix_eq!` bug found in the wild! This one is a very obvious one that should have been caught already, but alas, it wasn't. Discovered it as I was working on improving the LU decomposition (at least it was easy to find and spot).


AtheMathmo:
Awesome, thanks!

PR #96, opened 2016-11-25
Make swap_rows and swap_cols no-op if self-swap

### LINKED ISSUE #93: `swap_rows()` invokes undefined behavior
If you call

```rust
matrix.swap_rows(index, index);
```

I believe it currently invokes undefined behavior, as it will perform `mem::swap` on elements that reside in the same memory location. The same is currently not true for `swap_cols()`, as it uses `ptr::swap`.

Moreover, even if this might work with the current Rust compiler, it still performs a full swap when we could let it do a no-op. If `swap_rows(index, index)` was a no-op, we could simplify some calling code which currently explicitly checks that we're not swapping a row with itself (see for example the `lup_decomp` code). This is also true for `swap_cols()`.

I think a fix should at the very least only perform the swap if `a != b` in `swap_rows()` and `swap_cols()`. Even better if we can rewrite it to avoid unsafe code without a performance penalty (I don't know if this is doable).

Andlon:
Should fix #93. I don't think it's possible to avoid the `unsafe` blocks at the moment without measurable detriment to performance.

I also took the opportunity to clean up the examples a little, employing our relatively new macros.

### DIFF
--- a/src/matrix/slice.rs
+++ b/src/matrix/slice.rs
@@ -200,7 +200,7 @@ pub trait BaseMatrix<T>: Sized {
     /// out-of-bounds this function will panic.
     ///
     /// This function will never panic if the `Main` diagonal
-    /// offset is used. 
+    /// offset is used.
     fn iter_diag(&self, k: DiagOffset) -> Diagonal<T, Self> {
         let (diag_len, diag_start) = match k.into() {
             DiagOffset::Main => (min(self.rows(), self.cols()), 0),
@@ -749,7 +749,7 @@ pub trait BaseMatrix<T>: Sized {
                 format!("Vector size {0} != {1} Matrix column count.",
                         y.size(),
                         self.cols()));
-        
+
         forward_substitution(self, y)
     }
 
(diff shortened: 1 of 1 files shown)


AtheMathmo:
This looks good!

I want to investigate whether this will add any (albeit _tiny_) performance hit. Do you have any reference on why `mem::swap`ing the same location is bad (other than being a wasted op)?

I'm confident this will get merged as it stands now, just want to do a little due diligence for my own benefit :).

Andlon:
I found the Rust documentation a little lacking on this. However, in the documentation for [std::ptr::swap](https://doc.rust-lang.org/std/ptr/fn.swap.html), it says:

> Swaps the values at two mutable locations of the same type, without deinitializing either. They may overlap, unlike mem::swap which is otherwise equivalent.

What exactly happens if they overlap is not clear, but it does at least suggest that `mem::swap` does not allow overlapping segments of memory. Remember that `mem::swap` is _not_ an unsafe function, and so calling it with overlapping segments is impossible in safe Rust. Hence there are no guarantees that this function will work at all when you give it overlapping items.

Andlon:
As for performance, I did run the benchmarks before and after without any noticeable improvements. Ultimately, I've only added a single `a == b` equality comparison conditional, which should be negligible for all intents and purposes. Of course, it's even better if you look into it too :)

AtheMathmo:
Ah I see what you mean now. Yes it seems that we are indeed invoking undefined behaviour and so should definitely have this check.

Given that you've done a brief look already I doubt I'll find anything. I'll dig around for a short while tomorrow and then will get this in.

Thanks!

PR #97, opened 2016-11-26
Implement NumCast-based try_into

### LINKED ISSUE #79: Helper functions to convert between matrix types
Currently it is a bit of a headache to convert between types for matrices. For example say we have the following 2x2 matrix of integers:

```
[1, 2;
 3, 4]
```

And we want to convert this to a matrix of floats. To do this currently in rulinalg we need something like this:

```rust
use num::ToPrimitive;

let (n, m) = (mat.rows(), mat.cols());
let float_data = mat.into_iter().map(|x| x.to_f64()).collect::<Vec<_>>();
let float_mat = Matrix::new(n, m, float_data);
```

Instead of this we could leverage the [`NumCast`](http://rust-num.github.io/num/num/trait.NumCast.html) trait. This would allow us to convert between the primitives. We could add a new function like the following:

```rust
impl<T: ToPrimitive> Matrix<T> {
    pub fn try_into<U: NumCast>(self) -> Result<Matrix<U>, ...> {
        let (n, m) = (self.rows(), self.cols());
        let new_data = self.into_iter().map(|x| try!(U::from(x).ok_or(...))).collect::<Vec<_>>();
        Matrix::new(n, m, float_data)
    }
}
```

This will allow us to cast between the primitive types, or any custom types that can go between primitives using the `ToPrimitive` and `NumCast` traits. We can get a more complete coverage if the [`TryInto` and `TryFrom`](https://doc.rust-lang.org/std/convert/index.html) traits are stabilized and implemented between primitives.

Andlon:
Need this functionality for writing some interesting QuickCheck-based property tests for the LU Decomposition, so in the process I made an attempt to resolve #79.

As @AtheMathmo stated in the issue, the better long-term solution would be to implement `TryFrom/TryInto` once they are stabilized.

Since the functionality relies on `NumCast`, the tests only try to test a small sample of the possible conversions, and also corner cases like empty matrices. Please let me know if there are corner cases I should include!

### DIFF
--- a/src/convert.rs
+++ b/src/convert.rs
@@ -5,7 +5,9 @@
 
 use std::convert::From;
 
-use libnum::Zero;
+use libnum::{Zero, ToPrimitive, NumCast};
+
+use error::{Error, ErrorKind};
 
 use super::matrix::{DiagOffset, Matrix, MatrixSlice, MatrixSliceMut, BaseMatrix};
 use super::vector::Vector;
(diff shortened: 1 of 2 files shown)


AtheMathmo:
Great work! Thank you!

Issue #87, opened 2016-11-02
Returning MatrixSlice when requesting rows and columns


AtheMathmo:
This issue is to move discussion from #84 and #78 in one place. I'll fill out this issue properly soon, but primarily we want to discuss how best to handle the return values on the row and column access functions.

AtheMathmo:
I want to put a little more momentum into this to see if we can come to a decision. After thinking a little more I'm coming around more with @Andlon 's thinking. I expect that most of our users would prefer `get_row` and `iter_row` functions to return a `MatrixSlice` directly. This is also plays more nicely with added `get_col` and `iter_col` functions. I think that the `as_contiguous_data` function will be a fairly ugly but necessary addition which allows performance-focused users to get what they need in rare cases.

I want to run a few tests to check that we don't see any obvious regressions from this modification. If all looks good then hopefully we can get this change in and push in a lot of the other PRs which are waiting on 0.4.

AtheMathmo:
One more disadvantage of switching `iter_rows` to return `MatrixSlice`: we currently implement `FromIterator` for `&[T]`. I think we would need to change this to work with `MatrixSlice` as currently we have some nice `mat.iter_rows().skip(1).take(3).filter(...).collect::<Matrix<_>>()` things we can do.

Maybe we can play around with `Into` to get this to work nicely?

Andlon:
@AtheMathmo: I'm probably missing something basic here, but just based on what you said, isn't it just a matter of implementing `FromIterator<MatrixSlice> for Matrix`...?

AtheMathmo:
Yes that would work and is much easier :). Thanks!

Andlon:
A related issue: Does it actually make sense to have functions that return an `Option` for a given row/col index? I've been thinking on this a lot, and I really cannot come up with a single situation where I would attempt to access a row/col without already knowing that the index is within bounds.

Unless there are obvious situations where this is truly useful, I think the API would be a lot cleaner if it simply panicked when the index is out of bounds:

```rust
mat.row_mut(i) += 2.0;
```

vs.

```rust
mat.get_row_mut(i).unwrap() += 2.0;
```

(Here I've used the proposed feature where these functions return a `MatrixSlice`)

Andlon:
I also just realized an ambiguity in the `FromIterator<MatrixSlice>` suggestion.

Consider the following:

```rust
// Rows
let mat: Matrix<f64> = other_mat.row_iter().collect();

// Columns
[...]

PR #99, opened 2016-12-02
Updating get_row functions to return MatrixSlice

### LINKED ISSUE #85: Typo in get_row_unchecked docs
In the docs we write:

> Returns the row of a matrix at the given index without doing unbounds checking

Instead should read:

> Returns the row of a matrix at the given index without doing any bounds checking

AtheMathmo:
This is a very incomplete PR. I'm just putting in the request so I can track outstanding work more easily. Still to do:

- [x] Add column equivalents
- [x] Update row iterator
- [x] Rename functions to remove `get_` prefix. (Does this make the unchecked functions look weird?)
- [x] Explore custom types for `Row` and `Col`.
- [x] Benchmarks to compare performance
- [x] Basic evaluation of vectorization issues. (Do we need to keep casting to contiguous slices?)
- [x] Indexing for Row and Column types

There is some discussion in issue #87 . There are other ideas here which I would like to explore if I have time.

With this PR I'll try to work towards an agreeable API for these changes.

Also resolved #85 with this PR.

### DIFF
--- a/src/matrix/decomposition/svd.rs
+++ b/src/matrix/decomposition/svd.rs
@@ -326,7 +326,7 @@ mod tests {
             assert!(!row.iter().take(idx).any(|&x| x > 1e-10));
             assert!(!row.iter().skip(idx + 1).any(|&x| x > 1e-10));
             // Assert non-negativity of diagonal elements
-            assert!(row[idx] >= 0.0);
+            assert!(row.raw_slice()[idx] >= 0.0);
         }
 
         let recovered = u * b * v.transpose();
@@ -349,7 +349,7 @@ mod tests {
 
         let mut singular_triplets = u_transposed.iter_rows().zip(b.diag().into_iter()).zip(v_transposed.iter_rows())
             // chained zipping results in nested tuple. Flatten it.
-            .map(|((u_col, singular_value), v_col)| (Vector::new(u_col), singular_value, Vector::new(v_col)));
+            .map(|((u_col, singular_value), v_col)| (Vector::new(u_col.raw_slice()), singular_value, Vector::new(v_col.raw_slice())));
 
         assert!(singular_triplets.by_ref()
             // For a matrix M, each singular value σ and left and right singular vectors u and v respectively
--- a/src/matrix/decomposition/svd.rs
+++ b/src/matrix/decomposition/svd.rs
@@ -326,6 +326,7 @@ mod tests {
             assert!(!row.iter().take(idx).any(|&x| x > 1e-10));
             assert!(!row.iter().skip(idx + 1).any(|&x| x > 1e-10));
             // Assert non-negativity of diagonal elements
+            // TODO: Fix when indexing fixed
             assert!(row.raw_slice()[idx] >= 0.0);
         }
 
--- a/src/matrix/decomposition/svd.rs
+++ b/src/matrix/decomposition/svd.rs
@@ -327,7 +327,7 @@ mod tests {
             assert!(!row.iter().skip(idx + 1).any(|&x| x > 1e-10));
             // Assert non-negativity of diagonal elements
             // TODO: Fix when indexing fixed
-            assert!(row.raw_slice()[idx] >= 0.0);
+            assert!(row[idx] >= 0.0);
         }
 
         let recovered = u * b * v.transpose();
--- a/src/matrix/decomposition/svd.rs
+++ b/src/matrix/decomposition/svd.rs
@@ -326,7 +326,6 @@ mod tests {
             assert!(!row.iter().take(idx).any(|&x| x > 1e-10));
             assert!(!row.iter().skip(idx + 1).any(|&x| x > 1e-10));
             // Assert non-negativity of diagonal elements
-            // TODO: Fix when indexing fixed
             assert!(row[idx] >= 0.0);
         }
 
--- a/src/matrix/iter.rs
+++ b/src/matrix/iter.rs
@@ -86,6 +86,7 @@ impl<'a, T> Iterator for $rows<'a, T> {
             unsafe {
 // Get pointer and create a slice from raw parts
                 let ptr = self.slice_start.offset(self.row_pos as isize * self.row_stride);
+                // TODO: Remove after benching
                 //row = slice::$slice_from_parts(ptr, self.slice_cols);
                 row = $row_base {
                     row: $slice_base::from_raw_parts(ptr, 1, self.slice_cols, self.row_stride as usize)
@@ -105,6 +106,7 @@ impl<'a, T> Iterator for $rows<'a, T> {
             unsafe {
 // Get pointer to last row and create a slice from raw parts
                 let ptr = self.slice_start.offset((self.slice_rows - 1) as isize * self.row_stride);
+                // TODO: Remove after benching
                 //Some(slice::$slice_from_parts(ptr, self.slice_cols))
                 Some($row_base {
                     row: $slice_base::from_raw_parts(ptr, 1, self.slice_cols, self.row_stride as usize)
@@ -120,6 +122,7 @@ impl<'a, T> Iterator for $rows<'a, T> {
             let row: $row_type;
             unsafe {
                 let ptr = self.slice_start.offset((self.row_pos + n) as isize * self.row_stride);
+                // TODO: Remove after benching
                 //row = slice::$slice_from_parts(ptr, self.slice_cols);
                 row = $row_base {
                     row: $slice_base::from_raw_parts(ptr, 1, self.slice_cols, self.row_stride as usize)
@@ -144,6 +147,7 @@ impl<'a, T> Iterator for $rows<'a, T> {
     );
 );
 
+// TODO: Remove after benching
 // impl_iter_rows!(Rows, &'a [T], from_raw_parts);
 // impl_iter_rows!(RowsMut, &'a mut [T], from_raw_parts_mut);
 impl_iter_rows!(Rows, Row<'a, T>, Row, MatrixSlice);
@@ -184,6 +188,7 @@ impl<'a, T> ExactSizeIterator for RowsMut<'a, T> {}
 /// // where the first entry is less than 6.
 /// let b = a.iter_rows()
 ///          .skip(1)
+///          // TODO: Fix when indexing implemented
 ///          .filter(|x| x.raw_slice()[0] < 6)
 ///          .collect::<Matrix<usize>>();
 ///
--- a/src/matrix/iter.rs
+++ b/src/matrix/iter.rs
@@ -268,7 +268,7 @@ impl<'a, T: 'a + Copy> FromIterator<$row_type> for Matrix<T> {
         }
 
         for row in iterator {
-            assert!(row.row.cols() == cols, "Iterator slice length must be constant.");
+            assert!(row.row.cols() == cols, "Iterator row size must be constant.");
             mat_data.extend_from_slice(row.raw_slice());
             rows += 1;
         }
--- a/src/matrix/iter.rs
+++ b/src/matrix/iter.rs
@@ -2,7 +2,8 @@ use std::iter::{ExactSizeIterator, FromIterator};
 use std::mem;
 use std::slice;
 
-use super::{Matrix, MatrixSlice, MatrixSliceMut, Rows, RowsMut, Diagonal, DiagonalMut};
+use super::{Matrix, MatrixSlice, MatrixSliceMut};
+use super::{Row, RowMut, Rows, RowsMut, Diagonal, DiagonalMut};
 use super::slice::{BaseMatrix, BaseMatrixMut, SliceIter, SliceIterMut};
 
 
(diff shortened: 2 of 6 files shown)


AtheMathmo:
A reasonable vectorization test would be:

```rust
// With `iter_rows_mut` returning `&mut [T]`
for row in mat.iter_rows_mut() {
    for v in row {
       *v += 2;
   }
}

// Versus

for row in mat.iter_rows_mut() {
    row += 2;
}
```

AtheMathmo:
This PR is now ready for some initial reviews: @Andlon @andrewcsmith if either of you have time I would appreciate it!

I will update the top post above with more information soon. The gist is that I think I've settled on the API, but we still need some more quality of life changes. Some notable ones:

- `IntoIterator` for `Row` and `Column`
- `Index` for `Row`, etc.
- Doc improvements
- Some code reorganization, maybe another PR.
- Performance specialisations of `BaseMatrix` for `Row` and `Column`. Maybe another PR.

Those are the biggest right now (I think). I'll try to write a complete list up later.

src/matrix/iter.rs
```diff
@@ -89,7 +86,10 @@ impl<'a, T> Iterator for $rows<'a, T> {
             unsafe {
 // Get pointer and create a slice from raw parts
                 let ptr = self.slice_start.offset(self.row_pos as isize * self.row_stride);
-                row = slice::$slice_from_parts(ptr, self.slice_cols);
+                //row = slice::$slice_from_parts(ptr, self.slice_cols);
```
Andlon: Before final merge: remove this?

src/matrix/iter.rs
```diff
@@ -117,7 +120,10 @@ impl<'a, T> Iterator for $rows<'a, T> {
             let row: $row_type;
             unsafe {
                 let ptr = self.slice_start.offset((self.row_pos + n) as isize * self.row_stride);
-                row = slice::$slice_from_parts(ptr, self.slice_cols);
+                //row = slice::$slice_from_parts(ptr, self.slice_cols);
```
Andlon: Before final merge: remove this?

src/matrix/iter.rs
```diff
@@ -138,8 +144,10 @@ impl<'a, T> Iterator for $rows<'a, T> {
     );
 );
 
-impl_iter_rows!(Rows, &'a [T], from_raw_parts);
-impl_iter_rows!(RowsMut, &'a mut [T], from_raw_parts_mut);
+// impl_iter_rows!(Rows, &'a [T], from_raw_parts);
+// impl_iter_rows!(RowsMut, &'a mut [T], from_raw_parts_mut);
```
Andlon: Before final merge: remove this?

src/matrix/iter.rs
```diff
@@ -225,6 +233,57 @@ impl<'a, T: 'a + Copy> FromIterator<&'a [T]> for Matrix<T> {
...
+                    cols: 0,
+                }
+            }
+            Some(row) => {
+                rows += 1;
+                // Here we set the capacity - get iterator size and the cols
+                let (lower_rows, _) = iterator.size_hint();
+                cols = row.row.cols();
+
+                mat_data = Vec::with_capacity(lower_rows.saturating_add(1).saturating_mul(cols));
```
Andlon: Why saturating arithmetic here? I'm a bit confused about this part.

src/matrix/iter.rs
AtheMathmo: I left these here for easier benchmarking later. Just so I can quickly check if something is funky with this iterator.

src/matrix/iter.rs
```diff
@@ -225,6 +233,57 @@ impl<'a, T: 'a + Copy> FromIterator<&'a [T]> for Matrix<T> {
...
+                mat_data.extend_from_slice(row.raw_slice());
+            }
+        }
+
+        for row in iterator {
+            assert!(row.row.cols() == cols, "Iterator slice length must be constant.");
```
Andlon: The message could perhaps be a bit clearer for the user. For example, "Iterator of rows must have rows of compatible dimensions." or so?

src/matrix/iter.rs
```diff
@@ -105,7 +105,10 @@ impl<'a, T> Iterator for $rows<'a, T> {
             unsafe {
 // Get pointer to last row and create a slice from raw parts
                 let ptr = self.slice_start.offset((self.slice_rows - 1) as isize * self.row_stride);
-                Some(slice::$slice_from_parts(ptr, self.slice_cols))
+                //Some(slice::$slice_from_parts(ptr, self.slice_cols))
```
Andlon: Before final merge: Remove this?

src/matrix/mod.rs
```diff
@@ -73,6 +75,91 @@ pub struct MatrixSliceMut<'a, T: 'a> {
...
+/// of the row.
+///
+/// # Example
+///
+/// ```
+/// # #[macro_use] extern crate rulinalg; fn main() {
+/// use rulinalg::matrix::BaseMatrixMut;
+///
+/// let mut mat = matrix![1.0, 2.0;
+///                   3.0, 4.0];
```
Andlon: Minor detail: is the alignment off also in the rendered docs?

src/matrix/mod.rs
```diff
@@ -73,6 +75,91 @@ pub struct MatrixSliceMut<'a, T: 'a> {
...
+///
+/// {
+///     let mut row = mat.row_mut(1).unwrap();
```
Andlon: Can we remove the `Option` alltogether and just panic when index is out of bounds instead?

src/matrix/mod.rs
AtheMathmo: Thanks for pointing this out. It probably is.

src/matrix/iter.rs
AtheMathmo: Yes you're right. I overlooked this when I copied the existing `FromIterator` implementation. Also - do you think we should remove the `FromIterator<&'a [T]>` implementation? I think we probably should.

src/matrix/iter.rs
AtheMathmo: Thanks, will remove this soon!

src/matrix/iter.rs
AtheMathmo: That is a good question. I'm not sure what my initial motivation was but there is no _real_ reason.

AtheMathmo:
I think I answered all your questions except for:

> I've looked really hard, but I can't find any changes to `fn row(index)` on `BaseMatrix` anywhere, yet you seem to have changed this, according to the examples... Am I blind here?

This is because this function remained unchanged (except for the name and return type). The change was in `row_unchecked` which is called within this function.

Andlon:
I just realized that GitHub hadn't shown me all the changes! I had to press "Load Diff" at the bottom of the screen. Never experienced that before... Will take a look at the rest too

src/matrix/slice.rs
```diff
@@ -75,24 +76,83 @@ pub trait BaseMatrix<T>: Sized {
...
+    /// # Examples
+    ///
+    /// ```
+    /// # #[macro_use]
+    /// # extern crate rulinalg;
+    ///
+    /// # fn main() {
+    /// use rulinalg::matrix::{Matrix, BaseMatrix};
+    ///
+    /// let mat = Matrix::new(3,3, (0..9).collect::<Vec<usize>>());
```
Andlon: I know this is practical, but I personally find it easier to immediately understand examples if the example matrices are actually typed out when they are so small. It leaves the user to focus on the logic that's being showcased rather than having to parse the matrix construction logic. Not a big deal though!

src/matrix/slice.rs
```diff
@@ -887,48 +960,122 @@ pub trait BaseMatrixMut<T>: BaseMatrix<T> {
...
-            unsafe { Some(self.get_row_unchecked_mut(index)) }
+            unsafe { Some(self.row_unchecked_mut(index)) }
         } else {
             None
         }
     }
 
     /// Returns a mutable reference to the row of a matrix at the given index
-    /// without doing unbounds checking
+    /// without doing a bounds check.
```
Andlon: Haha, glad you got this one too. Let's remember to close #85 when you're done here ;)

src/matrix/slice.rs
AtheMathmo: No I agree, I'll make this change too.

AtheMathmo:
Benchcmp output:

```
 name                                           old ns/iter  new ns/iter  diff ns/iter  diff % 
 linalg::iter::empty                            0            0                       0    NaN% 
 linalg::iter::mat_diag_iter_100_500            220          221                     1   0.45% 
 linalg::iter::mat_diag_iter_10_50              73           76                      3   4.11% 
 linalg::iter::mat_diag_manual_100_500          304          303                    -1  -0.33% 
 linalg::iter::mat_diag_manual_10_50            66           66                      0   0.00% 
 linalg::matrix::empty                          0            0                       0    NaN% 
 linalg::matrix::mat_create_100_100             3,408        3,402                  -6  -0.18% 
 linalg::matrix::mat_create_add_100_100         10,122       10,168                 46   0.45% 
 linalg::matrix::mat_elediv_63_1000             83,484       83,765                281   0.34% 
 linalg::matrix::mat_elemul_63_1000             35,947       36,144                197   0.55% 
 linalg::matrix::mat_mat_elemul_63_1000         35,961       36,155                194   0.54% 
 linalg::matrix::mat_mul_10_10                  799          803                     4   0.50% 
 linalg::matrix::mat_mul_128_100                252,639      253,254               615   0.24% 
 linalg::matrix::mat_mul_128_1000               2,391,344    2,406,185          14,841   0.62% 
 linalg::matrix::mat_ref_add_100_100            8,564        8,520                 -44  -0.51% 
 linalg::matrix::mat_sum_rows_and_cols_128_100  12,245       12,286                 41   0.33% 
 linalg::matrix::mat_swap_cols_0_99             75           76                      1   1.33% 
 linalg::matrix::mat_swap_rows_0_99             18           18                      0   0.00% 
 linalg::svd::svd_100_100                       74,318,932   82,978,883      8,659,951  11.65% 
 linalg::svd::svd_10_10                         171,847      174,383             2,536   1.48%
```

The only non-noisy changes are SVD and diag iterator which should be unchanged with this. I'm not sure why they were so significantly affected but I'm not too concerned.

Andlon:
Looks great to me :)

I've noticed that the output of benchmarks can change _significantly_ unless I completely shut down all other programs running in the background (despite running on a quadcore i7 with low load). Guess it's due to the Linux scheduler and/or memory access patterns. Might account for the noise in the SVD benchmarks.

AtheMathmo:
> I've noticed that the output of benchmarks can change significantly unless I completely shut down all other programs running in the background (despite running on a quadcore i7 with low load).

I see similar things myself which is why I wasn't too concerned. I'll do some vectorization tests and if those look good then hopefully one more look through should make this merge-ready!

AtheMathmo:
I did a little digging into vectorization - for the most part it looks good to me. All of the operations break down into per-row operations. We may incur some cost for creating the row iterator unnecessarily (instead of the user iterating through the slice) but I don't think it will be significant.

Otherwise I can't find any cases where treating the row as a matrix instead of a `&[T]` would break vectorization. It's likely I've missed something but given the gain in utility here I'm happy to tick that box too.

AtheMathmo:
Ok I think I'm happy with this PR. Here are my thoughts on some outstanding things:

- `impl_ops` and `slice` module are very bloated. I'd like to fix this in a future PR.
- `IntoIterator` for `Row` and `Column`. I've decided against this as we get the expected behaviour by calling `iter()`. Additionally there are few cases where we can consume the row (for example we cannot in `iter_rows()`).
- Performance specializations for `Row` and `Col`. I'm not sure if anything is needed here really - it would help strictly but would have some code bloat so I think we should wait until this is an issue.

I'd love to hear from @Andlon , @andrewcsmith and anyone else to see if they spot anything else.

src/matrix/decomposition/svd.rs
```diff
@@ -326,7 +326,8 @@ mod tests {
             assert!(!row.iter().take(idx).any(|&x| x > 1e-10));
             assert!(!row.iter().skip(idx + 1).any(|&x| x > 1e-10));
             // Assert non-negativity of diagonal elements
-            assert!(row.raw_slice()[idx] >= 0.0);
+            // TODO: Fix when indexing fixed
+            assert!(row[idx] >= 0.0);
```
Andlon: I assume the TODO is out of date, and that this is actually fixed now?

Andlon:
> impl_ops and slice module are very bloated. I'd like to fix this in a future PR.

Good idea! I've also had some unpleasant run-ins with those modules.

> IntoIterator for Row and Column. I've decided against this as we get the expected behaviour by calling iter(). Additionally there are few cases where we can consume the row (for example we cannot in iter_rows()).

Makes sense.

> Performance specializations for Row and Col. I'm not sure if anything is needed here really - it would help strictly but would have some code bloat so I think we should wait until this is an issue.

Agree with your assessment. There are more important things (and possibly more interesting?) for us to spend time on, I think :)

src/matrix/decomposition/svd.rs
AtheMathmo: Ah yes! Forgot to remove it.. thanks

AtheMathmo:
I'm going to give this 12 hours in case @andrewcsmith wants to take a look then I'll be looking to merge. As there are breaking changes here I'm going to try to pick out the issues we should get in as part of 0.4 release. Ideally we wont have to wait too long before our next release.

AtheMathmo:
Merging now - mostly to force myself to finish up the other 0.4 issues!

PR #100, opened 2016-12-04
Add ability to select rows into pre-allocated buffer

theotherphil:
"into" is probably a bad name, given that this usually means that you're consuming self. Any better ideas?

Do you agree that this is a useful function to have? (It would for example remove a load of unnecessary allocations in cross validation.) Any opinions on whether the target should be a raw slice as here or something else (e.g. a matrix)?

I'll update the pull request to add a matching select_cols function when/if you're happy with the rows version.

### DIFF
--- a/src/matrix/slice.rs
+++ b/src/matrix/slice.rs
@@ -347,6 +347,51 @@ pub trait BaseMatrix<T>: Sized {
         }
     }
 
+    /// Select rows from matrix into pre-allocated buffer.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use rulinalg::matrix::{Matrix, BaseMatrix};
+    ///
+    /// let a = Matrix::new(3, 3, (1..10).collect::<Vec<usize>>());
+    ///
+    /// let mut b = vec![0; 3];
+    /// a.select_rows_into(&[2], &mut b);
+    /// assert_eq!(&b, &[7, 8, 9]);
+    ///
+    /// let mut c = vec![0; 6];
+    /// a.select_rows_into(&[1,2], &mut c);
+    /// assert_eq!(&c, &[4, 5, 6, 7, 8, 9]);
+    /// ```
+    ///
+    /// # Panics
+    ///
+    /// - Panics if row indices exceed the matrix dimensions.
+    /// - Panics if the target buffer has the wrong length.
+    fn select_rows_into<'a, I>(&self, rows: I, buffer: &mut [T])
+        where T: Copy,
+              I: IntoIterator<Item=&'a usize>,
+              I::IntoIter: ExactSizeIterator + Clone
+    {
+        let row_iter = rows.into_iter();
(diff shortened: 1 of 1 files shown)


Andlon:
I can definitely see the purpose of such an addition. However, I think with @AtheMathmo's proposed changes in #99, doing this will be very simple even without a helper function. Something like:

```rust
// Select rows from `mat` into existing matrix `buffer`
assert!(buffer.rows() == rows.len())
for i in rows {
     *buffer.row(i) = &mat.row(i);
}
```

At least, I think that should work, if I understand correctly what James has been working on. Hopefully he can clarify this.

theotherphil:
Ah, right. That sounds good!

Andlon:
@theotherphil: please see my second edit. It actually does not work out quite as nicely as I proposed there. I forgot that Rust does not allow you to "override" the assignment operator, only the arithmetic ones such as `+=`. In any case, it's perhaps a more flexible approach to just add a method on the new `RowMut` struct, or a method on a `MatrixSliceMut` such as `.replace(other)`. We should definitely think about this, because I agree that it's important to accommodate more optimized allocation patterns!

AtheMathmo:
@Andlon there is currently a `set_to` function on `BaseMatrixSlice` that does what you want. The code would look like this (I think):

```
// Select rows from `mat` into existing matrix `buffer`
assert!(buffer.rows() == rows.len())
for (target_row, source_row) in rows.enumerate() {
     buffer.row(target_row).set_to(*mat.row(source_row));
}
```

I think I am in agreement with @Andlon - hopefully the addition of the row type makes this a lot easier and we don't _need_ a helper function. Though this opinion is motivated by my belief that there probably are not too many use cases for such a function (however I agree it is a useful pattern to have access to).

All that said, I think this is a fairly neat function and if we can come up with some fairly common use cases it may be worth adding it.

theotherphil:
Ok, great. I'm not that fussed about the existence of the function itself - I just wanted an easy way of copying rows without resorting to copying elements individually. The code in @AtheMathmo's last comment looks good to me. I'll wait for his changes to land.

AtheMathmo:
> I'll wait for his changes to land.

Hopefully it wont take long! I'll leave this PR open in case it gets stalled, we can merge this as a temporary fix.

theotherphil:
Not directly relevant to this library, but possibly related to the changes you're currently working on: are there any plans to change the rusty-machine models to have input type MatrixSlice instead of Matrix?

AtheMathmo:
I have a [POC branch](https://github.com/AtheMathmo/rusty-machine/tree/poc-slice-training) where I was toying with this idea. I do want this and I think the upcoming rulinalg changes will make it worth the jump!

AtheMathmo:
I'm going to close this one out as the changes discussed above have landed in master. If there is ever future demand for this kind of function we can add it in fairly easily (hopefully).

PR #101, opened 2016-12-12
New metric traits

### LINKED ISSUE #81: Updates to Metric trait
The current [`Metric`](https://athemathmo.github.io/rulinalg/doc/rulinalg/trait.Metric.html) trait is pretty lackluster. It has a `norm` function which computes the euclidean norm. We could do a lot more to flesh it out. This issue is to open discussion about what this trait should support.

Some suggestions:

- Adding L-norms.
   - How should we define `L`, usize, float, enum?
   - How would the [L^infinity norm](http://mathworld.wolfram.com/L-Infinity-Norm.html) be defined?
- Should we directly allow distances (i.e. metrics...). Perhaps we should split out the norms into a new trait and use this trait in the definition of `Metric`: `pub trait Metric : Norm`.
- We could cover **a lot** here. Where should we define our limit?

In rusty-machine we often want to define a metric to be used as a prediction/classification error. For example we can choose the metric in k-means used to compute distance between elements in each set. It would be nice if we could provide a trait in rulinalg which we can be freely generic over in rulinalg.

Any work defined from this issue would be a breaking change - we should aim to include these in a 0.4 release.

AtheMathmo:
This PR attempts to resolve #81. New features include:

- Can define multiple norms using the `MatrixNorm` and `VectorNorm` traits.
- Automatically get the [_induced metric_ (fact 3)](https://www.math.usm.edu/perry/old_classes/mat681sp14/norm_and_metric.pdf) from the norm.

Some issues to work through:

- Should we keep a default euclidean implementation? _No_
- We have some performance hit for euclidean norm on `Matrix` - can specialize the implementation if we use a default function for this. _We will ignore this for now, but make a note for later._
- We have a set of traits each for vector and matrix types.

And some of the work left to do:

- [x] Add Lp norms.
- [x] Add documentation.
- [x] Add unit tests
- [x] Add a generic `norm` function to the `BaseMatrix` and `Vector` types so the user can specify which norm to use more easily.

---

Here are some benchmarks (without any Lp optimizations):

```
running 12 tests
test linalg::norm::lp_1_mat_100_50                 ... bench:       9,036 ns/iter (+/- 993)
test linalg::norm::lp_1_mat_10_50                  ... bench:         913 ns/iter (+/- 94)
test linalg::norm::sum_abs_mat_100_50              ... bench:       9,164 ns/iter (+/- 1,990)
test linalg::norm::sum_abs_mat_10_50               ... bench:         909 ns/iter (+/- 87)
test linalg::norm::lp_2_mat_100_50                 ... bench:      10,038 ns/iter (+/- 1,204)
test linalg::norm::lp_2_mat_10_50                  ... bench:       1,056 ns/iter (+/- 226)
test linalg::norm::euclidean_mat_100_50            ... bench:       3,266 ns/iter (+/- 392)
test linalg::norm::euclidean_mat_10_50             ... bench:         303 ns/iter (+/- 26)
test linalg::norm::lp_3_mat_100_50                 ... bench:     364,378 ns/iter (+/- 25,537)
test linalg::norm::lp_3_mat_10_50                  ... bench:      35,908 ns/iter (+/- 3,867)
test linalg::norm::sum_abs_powi_3_mat_100_50       ... bench:       9,436 ns/iter (+/- 1,171)
test linalg::norm::sum_abs_powi_3_mat_10_50        ... bench:         913 ns/iter (+/- 279)
```

### DIFF
--- a/src/norm/mod.rs
+++ b/src/norm/mod.rs
@@ -1,4 +1,27 @@
 //! The norm module
+//!
+//! This module contains implementations of various linear algebra norms.
+//! The implementations are contained within the `VectorNorm` and 
+//! `MatrixNorm` traits. This module also contains `VectorMetric` and
+//! `MatrixMetric` traits which are used to compute the metric distance.
+//!
+//! These traits can be used directly by importing implementors from
+//! this module. In most cases it will be easier to use the `norm` and
+//! `metric` functions which exist for both vectors and matrices. These
+//! functions take generic arguments for the norm to be used.
+//!
+//! In general you should use the least generic norm that fits your purpose.
+//! For example you would choose to use a `Euclidean` norm instead of an
+//! `Lp(2.0)` norm - despite them being mathematically equivalent. 
+//!
+//! # Defining your own norm
+//!
+//! Note that these traits enforce no requirements on the norm. It is up
+//! to the user to ensure that they define a norm correctly.
+//!
+//! To define your own norm you need to implement the `MatrixNorm`
+//! and/or the `VectorNorm` on your own struct. When you have defined
+//! a norm you get the _induced metric_ for free.
 
 use matrix::BaseMatrix;
 use vector::Vector;
@@ -34,7 +57,8 @@ pub trait MatrixMetric<'a, 'b, T, M1: 'a + BaseMatrix<T>, M2: 'b + BaseMatrix<T>
 /// The induced vector metric
 ///
 /// Given a norm `N`, the induced vector metric `M` computes
-/// the metric distance `d` as follows:
+/// the metric distance, `d`, between two vectors `v1` and `v2`
+/// as follows:
 ///
 /// `d = M(v1, v2) = N(v1 - v2)`
 impl<U, T> VectorMetric<T> for U
@@ -44,6 +68,13 @@ impl<U, T> VectorMetric<T> for U
     }
 }
 
+/// The induced matrix metric
+///
+/// Given a norm `N`, the induced matrix metric `M` computes
+/// the metric distance, `d`, between two matrices `m1` and `m2`
+/// as follows:
+///
+/// `d = M(m1, m2) = N(m1 - m2)`
 impl<'a, 'b, U, T, M1, M2> MatrixMetric<'a, 'b, T, M1, M2> for U
     where U: MatrixNorm<T, M1>,
     M1: 'a + BaseMatrix<T>,
@@ -56,6 +87,11 @@ impl<'a, 'b, U, T, M1, M2> MatrixMetric<'a, 'b, T, M1, M2> for U
 }
 
 /// The Euclidean norm
+///
+/// The Euclidean norm computes the square-root
+/// of the sum of squares.
+///
+/// `||v|| = SQRT(SUM(v_i * v_i))`
 #[derive(Debug)]
 pub struct Euclidean;
 
@@ -78,6 +114,17 @@ impl<T: Float, M: BaseMatrix<T>> MatrixNorm<T, M> for Euclidean {
 }
 
 /// The Lp norm
+///
+/// The Lp norm computes the `p`th root
+/// of the sum of elements to the `p`th power.
+///
+/// The Lp norm requires `p` to be greater than
+/// or equal `1`.
+///
+/// # p = infinity
+///
+/// In the special case where `p` is positive infinity,
+/// the Lp norm becomes a supremum over the absolute values.
 #[derive(Debug)]
 pub struct Lp<T: Float>(T);
 
--- a/src/norm/mod.rs
+++ b/src/norm/mod.rs
@@ -180,21 +180,56 @@ impl<T: Float, M: BaseMatrix<T>> MatrixNorm<T, M> for Lp<T> {
 mod tests {
     use super::*;
     use vector::Vector;
-    use matrix::MatrixSlice;
+    use matrix::{Matrix, MatrixSlice};
 
     #[test]
-    fn test_euclidean_vector() {
+    fn test_euclidean_vector_norm() {
         let v = Vector::new(vec![3.0, 4.0]);
         assert!((VectorNorm::norm(&Euclidean, &v) - 5.0) < 1e-30);
     }
 
     #[test]
-    fn test_euclidean_matrix() {
+    fn test_euclidean_matrix_norm() {
         let m = matrix![3.0, 4.0;
                         1.0, 3.0];
         assert!((MatrixNorm::norm(&Euclidean, &m) - 35.0.sqrt()) < 1e-30);
 
         let slice = MatrixSlice::from_matrix(&m, [0,0], 1, 2);
         assert!((MatrixNorm::norm(&Euclidean, &slice) - 5.0) < 1e-30);
+    }
+
+    #[test]
+    fn test_euclidean_vector_metric() {
+        let v = Vector::new(vec![3.0, 4.0]);
+        assert!((VectorMetric::metric(&Euclidean, &v, &v)) < 1e-30);
+
+        let v1 = Vector::new(vec![0.0, 0.0]);
+        assert!((VectorMetric::metric(&Euclidean, &v, &v1) - 5.0) < 1e-30);
+
+        let v2 = Vector::new(vec![4.0, 3.0]);
+        assert!((VectorMetric::metric(&Euclidean, &v, &v2) - 2.0.sqrt()) < 1e-30);
+    }
+
+    #[test]
+    fn test_euclidean_matrix_metric() {
+        let m = matrix![3.0, 4.0;
+                        1.0, 3.0];
+        assert!((MatrixMetric::metric(&Euclidean, &m, &m)) < 1e-30);
+
+        let m1 = Matrix::zeros(2, 2);
+        assert!((MatrixMetric::metric(&Euclidean, &m, &m1) - 35.0.sqrt()) < 1e-30);
+
+        let m2 = matrix![2.0, 3.0;
+                         2.0, 4.0];
+        assert!((MatrixMetric::metric(&Euclidean, &m, &m2) - 2.0) < 1e-30);
+    }
+
+    #[test]
+    #[should_panic]
+    fn test_euclidean_matrix_metric_bad_dim() {
+        let m = matrix![3.0, 4.0];
+        let m2 = matrix![1.0, 2.0, 3.0];
+
+        MatrixMetric::metric(&Euclidean, &m, &m2);
     }
 }
--- a/src/norm/mod.rs
+++ b/src/norm/mod.rs
@@ -131,73 +131,84 @@ impl<T: Float, M: BaseMatrix<T>> MatrixNorm<T, M> for Euclidean {
 ///
 /// In the special case where `p` is positive infinity,
 /// the Lp norm becomes a supremum over the absolute values.
+///
+/// TODO: Explain enum a little more
 #[derive(Debug)]
-pub struct Lp<T: Float>(pub T);
-
-impl<T: Float> Lp<T> {
-    /// Returns the L_infinity norm.
-    ///
-    /// # Examples
-    ///
-    /// ```
-    /// # #[macro_use] extern crate rulinalg; fn main() {
-    /// use rulinalg::norm::{MatrixNorm, Lp};
-    ///
-    /// let l_inf = Lp::infinity();
-    /// let mat = matrix![ 1.0, 2.0;
-    ///                   -4.0, 1.5];
-    /// assert_eq!(l_inf.norm(&mat), 4.0);
-    /// # }
-    /// ```
-    pub fn infinity() -> Lp<T> {
-        Lp(T::infinity())
-    }
+pub enum Lp<T: Float> {
+    /// The L-infinity norm (supremum)
+    Infinity,
+    /// The Lp norm where p is an integer
+    Integer(i32),
+    /// The Lp norm where p is a float
+    Float(T)
 }
 
 impl<T: Float> VectorNorm<T> for Lp<T> {
     fn norm(&self, v: &Vector<T>) -> T {
-        if self.0 < T::one() {
-            panic!("p value in Lp norm must >= 1")
-        } else if self.0.is_infinite() {
-            // Compute supremum
-            let mut abs_sup = T::zero();
-            for d in v.iter().map(|d| d.abs()) {
-                if d > abs_sup {
-                    abs_sup = d;
+        match *self {
+            Lp::Infinity => {
+                // Compute supremum
+                let mut abs_sup = T::zero();
+                for d in v.iter().map(|d| d.abs()) {
+                    if d > abs_sup {
+                        abs_sup = d;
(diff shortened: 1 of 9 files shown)


Andlon:
Looks interesting from a cursory look!

> Add a generic norm function to the BaseMatrix and Vector types so the user can specify which norm to use more easily.

Do you mean something like `mat.norm<Euclidean>()`? I think this is a very good idea! Then we're explicit about which norm is being used. I can't keep track of how many times I've had to look up functions like `.norm()` in other libraries/languages to make sure they are computing what I think they do. Should greatly increase readability.

AtheMathmo:
The function would have to look more like `mat.norm(Euclidean)`. The reason why is that I figured some norms require some data to be stored about them, `LpNorm` for example. Because of that all of the trait functions work on `&self`. I think if we can still keep the syntax fairly clean with this though - for example using tuple structs so that the syntax would be `mat.norm(Lp(2.0))`.

Do you think it is worth maintaining a separate function for the euclidean norm? I'm thinking that it probably isn't right now but I am bothered by the incurred performance hit. For clarity we lose performance as the `Euclidean` norm acts on a `BaseMatrix` so we can't assume the whole data block is contiguous.

Andlon:
I think unless the performance hit is very, very substantial (and I suspect computing norms is rarely a bottleneck of your computation in any case?), my gut feeling is that we should focus on making the API as easy-to-use and consistent as possible for now. Later, when i.e. specialization arrives, we can hopefully recover any lost performance. I think `rulinalg` is still so young that we can sacrifice some maximum performance for the sake of a better API, provided that these things can be remedied as Rust gains more features.

Also, note that you will run into potentially fairly hefty performance penalties by specifying p-norms at runtime: for example, if `p` is a float, then `Lp(2.0)` would have to perform a large amounts of fractional exponential operations instead of just simple elementwise multiplication `x * x`. That is, assuming that the compiler isn't extremely smart (I doubt it will recognize that exponentiating `x` to the power of `2.0` in floating-point is equivalent to `x * x`?). I'm not sure how to remedy this, however...

AtheMathmo:
The performance hit will be due to poor vectorization. Instead of vectorizing the dot operation over the whole matrix `Vec` we can only vectorize in chunks over the row. This shouldn't be too significant.

As for the `Lp` performance penalty, I think this is acceptable as we explicitly provide (somewhat) more efficient implementations with the `Euclidean` norm.

AtheMathmo:
I have the following left to do:

- Unit tests for the `Euclidean` and `Lp` norms.
- Documentation for the `norm` module. I want to specifically talk about the assumptions made by the traits and users elsewhere in rulinalg. This should help if somebody wants to implement a custom metric or norm.

AtheMathmo:
I think this is ready for a proper review now. I'll try to go over it all again but I'm fairly happy with it.

There are things I'd hope to change if new rust features land but otherwise I think this is a good start.

AtheMathmo:
I've had another look through and I am happy to merge this now. I thought I'd ping @Andlon in case you want to take one more look through. If you don't have time then don't worry! I'm just trying to rush through some of these so we can get the 0.4 release out.

Andlon:
Just about getting on a plane. I might have some time to look into it
tonight or tomorrow morning, if you can wait that long. If not then please
go ahead and merge :-)

AtheMathmo:
I can happily wait a couple of days. In the mean time I'm working on refactoring the `impl_ops` and `slice` modules. Thank you!

src/norm/mod.rs
```diff
@@ -0,0 +1,291 @@
...
+//! `Lp(2.0)` norm - despite them being mathematically equivalent. 
+//!
+//! # Defining your own norm
+//!
+//! Note that these traits enforce no requirements on the norm. It is up
+//! to the user to ensure that they define a norm correctly.
+//!
+//! To define your own norm you need to implement the `MatrixNorm`
+//! and/or the `VectorNorm` on your own struct. When you have defined
+//! a norm you get the _induced metric_ for free.
```
Andlon: Mathematically the induced metric is very clear, but perhaps just spend one additional sentence on explaining what this means in Rust terms? I.e. that anything that implements `MatrixNorm`/`VectorNorm` also gets an automatic `MatrixMetric`/`VectorMetric`?

src/norm/mod.rs
```diff
@@ -0,0 +1,291 @@
...
+/// The Lp norm computes the `p`th root
+/// of the sum of elements to the `p`th power.
+///
+/// The Lp norm requires `p` to be greater than
+/// or equal `1`.
+///
+/// # p = infinity
+///
+/// In the special case where `p` is positive infinity,
+/// the Lp norm becomes a supremum over the absolute values.
```
Andlon: Since math formulas are very hard to present in Github Markdown, perhaps we could link to an appropriate Wikipedia section which discusses the issue? I.e. [such as this](https://en.wikipedia.org/wiki/Lp_space#The_p-norm_in_finite_dimensions)

src/norm/mod.rs
```diff
@@ -0,0 +1,291 @@
...
+                }
+            }
+            abs_sup
+        } else {
+            // Compute standard lp norm
+            let mut s = T::zero();
+            for x in v {
+                s = s + x.abs().powf(self.0);
+            }
+            s.powf(self.0.recip())
```
Andlon: For now I think this is perfectly fine, but in the future we may want to investigate whether we should implement compensated summation such as [Kahan summation](https://en.wikipedia.org/wiki/Kahan_summation_algorithm) to improve accuracy due to rounding errors. I believe norm computations are very easily afflicted by such issues.

While this would require more floating point operations, it may not necessarily need to decrease performance so much since it's possibly that one is cache-bound in the first place.

src/norm/mod.rs
```diff
@@ -0,0 +1,291 @@
...
+    use std::f64;
+
+    use super::*;
+    use vector::Vector;
+    use matrix::{Matrix, MatrixSlice};
+
+    #[test]
+    fn test_euclidean_vector_norm() {
+        let v = Vector::new(vec![3.0, 4.0]);
+        assert!((VectorNorm::norm(&Euclidean, &v) - 5.0) < 1e-30);
```
Andlon: Note that while this arbitrary tolerance `1e-30` works on your machine (and Travis), it _may_ possibly give different results for different architectures (and as such fail on some other contributor's machine). I don't have a very good drop-in remedy for this, other than lowering it to something like `1e-14` or so, which also has its own problems.

src/norm/mod.rs
```diff
@@ -0,0 +1,291 @@
...
+        let m2 = matrix![1.0, 2.0, 3.0];
+
+        MatrixMetric::metric(&Euclidean, &m, &m2);
+    }
+
+    #[test]
+    fn test_lp_vector_supremum() {
+        let v = Vector::new(vec![-5.0, 3.0]);
+
+        let sup = VectorNorm::norm(&Lp(f64::INFINITY), &v);
```
Andlon: Perhaps also introduce static methods `Lp::supremum()` and `Lp::infimum()`? Or `max`/`min`, haven't thought this through very much.

Andlon: Eh, forget about infimum/min! (But might still be useful for supremum)

src/norm/mod.rs
AtheMathmo: I agree! Will add this before merging.

src/norm/mod.rs
AtheMathmo: Yes this is probably a good idea, thanks.

src/norm/mod.rs
AtheMathmo: This might be worth doing - we could add a `utils` function for Kahan summation.

I've never used it myself before and I'm inclined to delay this to a later PR (my only reason for not doing so would be the fear that it would be forgotten). I think it is at least worth exploring whether this gives us any accuracy improvements.

src/norm/mod.rs
AtheMathmo: Yes I agree - a lower tolerance is probably the better of two evils here.

src/norm/mod.rs
AtheMathmo: Yes this is also a good idea! I was toying with having `Lp` be an enum:

```rust
pub enum Lp<T> {
    Supremum,
    Zero,
    Normal(T),
    // .. Other special cases
}
```

Or something like that but decided against it for now. Do you think this would be a worthwhile change?

Either way, adding these helper functions may push me into splitting each norm into it's own sub-module. (Which is not a bad thing).

AtheMathmo:
Thank you for the review. There are definitely some things you pointed out that I would like to change before merging.

As for the runtime handling of special cases - I think this could be worthwhile. I'll write up some benchmarks before this gets merged and we can get an idea of how big the difference is. If it is a big performance hit then I think adding these special case handlers is a good idea.

Note that by using the `Lp` enum idea that I suggested in an above comment we can also alleviate some of this issue. For example, having an `Integer(i32)` variant would let us use the [`f64::powi`](https://doc.rust-lang.org/std/primitive.f64.html#method.powi) which would be faster.

AtheMathmo:
I added some benchmarks, here are the results:

```
running 12 tests
test linalg::norm::lp_1_mat_100_50                 ... bench:       9,036 ns/iter (+/- 993)
test linalg::norm::lp_1_mat_10_50                  ... bench:         913 ns/iter (+/- 94)
test linalg::norm::sum_abs_mat_100_50              ... bench:       9,164 ns/iter (+/- 1,990)
test linalg::norm::sum_abs_mat_10_50               ... bench:         909 ns/iter (+/- 87)
test linalg::norm::lp_2_mat_100_50                 ... bench:      10,038 ns/iter (+/- 1,204)
test linalg::norm::lp_2_mat_10_50                  ... bench:       1,056 ns/iter (+/- 226)
test linalg::norm::euclidean_mat_100_50            ... bench:       3,266 ns/iter (+/- 392)
test linalg::norm::euclidean_mat_10_50             ... bench:         303 ns/iter (+/- 26)
test linalg::norm::lp_3_mat_100_50                 ... bench:     364,378 ns/iter (+/- 25,537)
test linalg::norm::lp_3_mat_10_50                  ... bench:      35,908 ns/iter (+/- 3,867)
test linalg::norm::sum_abs_powi_3_mat_100_50       ... bench:       9,436 ns/iter (+/- 1,171)
test linalg::norm::sum_abs_powi_3_mat_10_50        ... bench:         913 ns/iter (+/- 279)
```

The summary:

- We don't need to specialize for `Lp(1.0)`.
- Specializing for p=2 gives us an ~3 times speedup.
- Specializing for integers (at compile time) gives us a huge speedup.

I'll copy these benchmarks into the PR description. I think that with these results it is worth using the enum approach for the Lp norm. Something like:

```rust
/// The Lp norm.
///
/// Usual stuff...
///
/// We use an enum here to allow performance specializations.
pub enum Lp<T: Float> {
    /// The L-infinity norm (supremum)
    pub Infinity,
    /// The Lp norm where p is an integer
    pub Integer(i32),
    /// The Lp norm where p is a float
    pub  Float(T)
}
```

Andlon:
The benchmarks are very interesting! I definitely agree that this warrants a solution which gives the compiler more information to work with, and I think your `enum` approach seems like a reasonable way to resolve this.

Did you by any chance check the `powi` approach vs `Euclidean`?

Considering that `p = 3` with `powi` takes about 3 times as much time as `Euclidean` (which of course is cheaper since it's `p = 2`), I wonder if the absolute value might be responsible for a considerable part of this extra runtime. That _might_ warrant specializing even-valued `p` so that they avoid the redundant `abs` calls.

In any case, I think the PR looks really good (especially with the enum if you want to go for that), so we can leave micro-optimizations for later work if you'd rather work on more pressing matters. At least the enum approach opens up for optimization work without further breaking changes to the API.

AtheMathmo:
I just checked the `powi` approach with `p = 2` and got numbers very similar to `p = 3`. I was hoping it would be slower than `Euclidean` as we put work into vectorizing the `Euclidean` approach. I'm not sure why it doesn't have a significant improvement over `p = 3` but I wont trouble myself with it...

I'm going to go with the enum approach. It seems like the best of both - and as you say I like that we can add more special cases later without breaking the API.

Andlon:
Yeah, for now there are bigger fish to fry. Better leave the micro-optimizations for later :)

(Though, in the future, we could perhaps also get rid of Euclidean by special-casing `p == 2` in `Lp`, I think. It's just a little unfortunate to have to maintain two pieces of functionality which amount to the same thing, especially if we want to introduce more elaborate summation etc. at a later point)

AtheMathmo:
I'm seeing something a little odd with the benchmarks after updating:

```
running 14 tests
test linalg::norm::lp_1_mat_100_50                 ... bench:      24,530 ns/iter (+/- 3,244)
test linalg::norm::lp_1_mat_10_50                  ... bench:       2,271 ns/iter (+/- 198)
test linalg::norm::sum_abs_mat_100_50              ... bench:       9,035 ns/iter (+/- 103)
test linalg::norm::sum_abs_mat_10_50               ... bench:       1,495 ns/iter (+/- 131)
test linalg::norm::euclidean_mat_100_50            ... bench:       3,222 ns/iter (+/- 72)
test linalg::norm::euclidean_mat_10_50             ... bench:         310 ns/iter (+/- 26)
test linalg::norm::lp_2_mat_100_50                 ... bench:      27,139 ns/iter (+/- 422)
test linalg::norm::lp_2_mat_10_50                  ... bench:       2,720 ns/iter (+/- 151)
test linalg::norm::lp_float_2_mat_100_50           ... bench:      60,295 ns/iter (+/- 1,848)
test linalg::norm::lp_float_2_mat_10_50            ... bench:       6,109 ns/iter (+/- 339)
test linalg::norm::lp_3_mat_100_50                 ... bench:      25,657 ns/iter (+/- 2,252)
test linalg::norm::lp_3_mat_10_50                  ... bench:       2,587 ns/iter (+/- 226)
test linalg::norm::lp_float_3_mat_100_50           ... bench:     367,841 ns/iter (+/- 32,373)
test linalg::norm::lp_float_3_mat_10_50            ... bench:      36,359 ns/iter (+/- 3,169)
```

Most of these numbers look fine - the only questionable one is the L1 norm. We would expect it to be very close to the `sum_abs` benchmarks. In fact I even tried modifying the L1 norm so that it _is_ exactly the same but still the difference remains. All I can think of is that the `sub_abs` test works on f64 instead of generics and maybe this allows some vectorization?

This isn't a huge concern to me and I'm still happy to merge. I thought I'd post it here in case you ( @Andlon ) can think of any reasons for this?

Andlon:
I can't explain why the difference remains even though you used the exact same code. Perhaps try clearing your `target` directory? Might be some cache corruption issue?

Also, by updating, do you mean updating your nightly version...? It's possible there's a regression on nightly. I guess the way to find out would be to go backwards through different nightly versions (or something like bisection), but I think maybe your time is better spent elsewhere, to be honest.

AtheMathmo:
I meant updating the PR. I'll try clearing the cache - if it still doesn't make a difference I wont worry about it too much. Benchmarking can be a little tempermental.

PR #103, opened 2016-12-16
API: impl FromIter for Vector

sinhrks:
For easier ``Vector`` creation.


AtheMathmo:
Thank you!

I'll merge this in soon as the code looks good. Just so you're aware, this may take a little while to be released. We have some breaking changes pending and I want to bundle a few more in before we do the 0.4 release. I should get through most of it this weekend.

sinhrks:
thx, np:)

PR #104, opened 2016-12-16
API: add get_unchecked to Vector

sinhrks:
Don't have to write ``vector.data().get_unchecked...`` to skip index boundary check.

### DIFF
--- a/src/vector.rs
+++ b/src/vector.rs
@@ -73,6 +73,18 @@ impl<T> Vector<T> {
         self.mut_data().iter_mut()
     }
 
+    /// Returns a pointer to the element at the given index, without doing
+    /// bounds checking. So use it very carefully!
+    pub unsafe fn get_unchecked(&self, index: usize) -> &T {
+        self.data().get_unchecked(index)
+    }
+
+    /// Returns an unsafe mutable pointer to the element in index.
+    /// So use it very carefully!
+    pub unsafe fn get_unchecked_mut(&mut self, index: usize) -> &mut T {
+        self.mut_data().get_unchecked_mut(index)
+    }
+
 }
 
 impl<T> Into<Vec<T>> for Vector<T> {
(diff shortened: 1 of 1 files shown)


src/vector.rs
```diff
@@ -73,6 +73,18 @@ impl<T> Vector<T> {
         self.mut_data().iter_mut()
     }
 
+    /// Returns a pointer to the element at the given index, without doing
+    /// bounds checking. So use it very carefully!
+    pub unsafe fn get_unchecked(&self, index: usize) -> &T {
```
AtheMathmo: Fairly trivial but you don't need to do `self.data()`, `self.data` is enough.

I'm sure the compiler optimizes this out but we should be explicit that this is what we want.

AtheMathmo:
Very minor point, otherwise this looks good.

Thank you again!

src/vector.rs
sinhrks: You're right. updated.

AtheMathmo:
Looks merge ready to me. Will give 24 hours for other comments then merge.

src/vector.rs
```diff
@@ -73,6 +73,18 @@ impl<T> Vector<T> {
...
+        self.data.get_unchecked(index)
+    }
+
+    /// Returns an unsafe mutable pointer to the element in index.
```
AtheMathmo: Was just about to merge this and noticed some things in the docs. They are minor but I'd like to see them changed before merged.

- Here `element in index` does not makes sense. Use `element at given index` as in other function.
- Similar to above, the descriptions should be very similar on both functions. The only difference being that one has `mutable pointer` where the other has `pointer`.
- I would remove the `So use it very carefully!`

AtheMathmo:
I made the tweaks myself as they were very small. I should also apologize as I see that you had just mirrored the docs in the standard library. Nonetheless I did change them a little to appease my own preferences.

sinhrks:
thx for the change and merge:)

PR #106, opened 2016-12-17
API: Vector::from_fn

sinhrks:
``Vector`` version of ``Matrix::from_fn``.

### DIFF
--- a/src/vector.rs
+++ b/src/vector.rs
@@ -47,8 +47,6 @@ impl<T> Vector<T> {
     /// and constructs a new vector such that `V_i = f(i)`,
     /// where `i` is the index.
     ///
-    /// Requires the vector data.
-    ///
     /// # Examples
     ///
     /// ```
@@ -850,7 +848,7 @@ mod tests {
     }
 
     #[test]
-    fn create_vector_new_from_fn() {
+    fn create_vector_from_fn() {
         let v1 = Vector::from_fn(3, |x| x + 1);
         assert_eq!(v1, Vector::new(vec![1, 2, 3]));
 
--- a/src/vector.rs
+++ b/src/vector.rs
@@ -43,6 +43,31 @@ impl<T> Vector<T> {
         }
     }
 
+    /// Constructor for Vector struct that takes a function `f`
+    /// and constructs a new vector such that `V_i = f(i)`,
+    /// where `i` is the index.
+    ///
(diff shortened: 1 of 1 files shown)


AtheMathmo:
Good idea! It's good to see the `vector` module getting some love.

I think this PR looks ready to merge. I'll hang on a little to see if anyone else has any comments.

Thanks.

src/vector.rs
```diff
@@ -43,6 +43,31 @@ impl<T> Vector<T> {
         }
     }
 
+    /// Constructor for Vector struct that takes a function `f`
+    /// and constructs a new vector such that `V_i = f(i)`,
+    /// where `i` is the index.
+    ///
+    /// Requires the vector data.
+    ///
```
theotherphil: Nitpick: I don't know what "Requires the vector data" means - I think this sentence should just be removed.

src/vector.rs
```diff
@@ -825,6 +850,25 @@ mod tests {
     }
 
     #[test]
+    fn create_vector_new_from_fn() {
+        let v1 = Vector::from_fn(3, |x| x + 1);
```
theotherphil: Even nitpickier - should this be called create_vector_from_fn?

src/vector.rs
AtheMathmo: I agree with this comment. In fact the advantage of this function is that it **does_not** require the vector data.

src/vector.rs
AtheMathmo: I think `from_fn` is better for a few reasons.

- We already have a `from_fn` for matrices
- It is shorter
- It starts with `from` which is conventional
- There is some precedence for this function. There used to be a `from_fn` for `std::Vec`, and it does exist [elsewhere in the ecosystem](http://www.piston.rs/image/image/struct.ImageBuffer.html#method.from_fn)

AtheMathmo:
Thank you for reviewing @theotherphil !

src/vector.rs
theotherphil: I meant the name of the test!

AtheMathmo: oh, ha! In which case I agree :D

sinhrks:
Thx, fixed the points.

AtheMathmo:
Looks merge ready to me. Will give 24 hours for other comments then merge.

PR #107, opened 2016-12-17
Removing self clone on referenced functions

### LINKED ISSUE #91: LU Decomposition should consume self
Currently we do `self.clone()` at the start. We should instead consume self to pass this choice to the user.

See #90

AtheMathmo:
This PR is part of our series of breaking change updates coming with 0.4.

It is bad practice to clone `self` in a function taking `&self`. This PR resolves #91 and similar situations.

### DIFF
--- a/src/matrix/decomposition/eigen.rs
+++ b/src/matrix/decomposition/eigen.rs
@@ -76,13 +76,13 @@ impl<T: Any + Float + Signed> Matrix<T> {
 
     }
 
-    fn francis_shift_eigenvalues(&self) -> Result<Vec<T>, Error> {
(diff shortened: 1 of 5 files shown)


AtheMathmo:
Here is a list of the exposed functions which now consume `self`:

- `mat.eigenvalues()`
- `mat.eigendecomp()`
- `mat.lup_decomp()`
- `mat.solve(x)`
- `mat.inverse()`
- `mat.det()`

It's quite a hefty list and I think people may question needing to consume the matrix for some of these operations. However, I think it is better to have any shortcomings of the library very visible. Hopefully we can provide alternatives in the future that are less efficient but do not require cloning or consuming the matrix.

PR #111, opened 2016-12-20
Splitting out the slice module

AtheMathmo:
In this PR I've shuffled around some of the larger modules. In particular I've tackled the `slice` module and the `matrix/mod` module. I've tried to complete the following changes:

- New `base` module for the `BaseMatrix` traits.
- Moved `BaseMatrix` implementations into new sub module
- Moved `impl Matrix` blocks to new module

This means that the `matrix/mod` module contains only declarations of structs that we use throughout the matrix module. The slice module contains only `impl MatrixSlice(Mut)` blocks.

The `base` mod is still fairly large but there is not much that can be done about this. The `BaseMatrix` traits are just pretty huge.


AtheMathmo:
Closing due to merge conflict

PR #112, opened 2016-12-22
 Rename iterator functions to *_iter

c410-f3r:
Please, disconsider PR #78.


AtheMathmo:
This looks good to me but I want to check something with others. Do we want `row_iter` or `rows_iter`?

My vote is for `row_iter` but I remember @tafia preferring `rows_iter`. Does anybody else prefer `rows_iter`?

Andlon:
I also prefer `row_iter`, but I'm not able to come up with a very rational justification as to why. It just... feels better.

AtheMathmo:
Ok, in that case we'll go for this as-is. I need to actually read through to make sure there's no tiny bugs. After I've done that I'll merge and we will be very close to a 0.4 release!

Thanks @c410-f3r

PR #114, opened 2016-12-23
Splitting out the slice module

AtheMathmo:
This PR replaces #111 - the merge conflict was too tricky.

Tomorrow morning I'll check through to make sure that I haven't accidentally dropped any unit tests. Then I'll merge this and get a PR in for the 0.4 release.

----

Copied from previous PR:

In this PR I've shuffled around some of the larger modules. In particular I've tackled the `slice` module and the `matrix/mod` module. I've tried to complete the following changes:

- New `base` module for the `BaseMatrix` traits.
- Moved `BaseMatrix` implementations into new sub module
- Moved `impl Matrix` blocks to new module

This means that the `matrix/mod` module contains only declarations of structs that we use throughout the matrix module. The slice module contains only `impl MatrixSlice(Mut)` blocks.

The `base` mod is still fairly large but there is not much that can be done about this. The `BaseMatrix` traits are just pretty huge.


AtheMathmo:
I'd appreciate if someone else could read the description above and just confirm that this sounds sensible. The code doesn't _need_ a proper review as I only moved things around and fixed the imports.

Andlon:
Yes, it sounds sensible :) Not sure if i.e. the `base` module needs to be public though, or whether it should just be a private implementation detail for code organization purposes?

AtheMathmo:
Thanks for the comment - I was being a little impatient and merged already (sorry!).

I _think_ `base` needs to be public as the documentation will redirect there. I remember checking a while ago but maybe I should check again. It would be worth hiding it if possible.

Andlon:
I think it can be private, but then you'd need to export its members from within a different module. For example, if you wanted `BaseMatrix` to be part of the `matrix` module, you could let the `base` module be private, and let the `matrix` module contain

```rust
mod base;
pub use self::base::BaseMatrix;
```

At least, I was under the impression that this would give the expected results, i.e. that `BaseMatrix` would be part of `matrix`, and the docs would also be tied to `matrix`. Does it not? (I don't know, I've just assumed it to be the case)

AtheMathmo:
I just checked and you are correct!

I agree that this is an obvious improvement. I'll push the one-line change directly to master.

PR #115, opened 2016-12-24
Add Matrix.min / max

### LINKED ISSUE #113: Matrix Row / column wise min/max
It is convenient if we have row / column-wise ``.min`` and ``.max`` of ``Matrix``. 

- accept ``Axes`` to specify axis like other aggregation funcs.
- min / max logic should be the same as ``argmin / argmax``. (thus requires ``PartialEq``)

sinhrks:
Closes #113.

Following is benchmark results. 

```
test linalg::matrix::mat_min_col                   ... bench:      15,455 ns/iter (+/- 1,490)
test linalg::matrix::mat_min_row                   ... bench:      31,269 ns/iter (+/- 161)
```

### DIFF
--- a/src/matrix/base/mod.rs
+++ b/src/matrix/base/mod.rs
@@ -461,7 +461,7 @@ pub trait BaseMatrix<T>: Sized {
         }
     }
 
-    /// The min of the specified axis of the matrix.
+    /// The max of the specified axis of the matrix.
     ///
     /// # Examples
     ///
@@ -474,10 +474,10 @@ pub trait BaseMatrix<T>: Sized {
     ///                 3.0, 4.0];
     ///
     /// let cmax = a.max(Axes::Col);
-    /// assert_eq!(cmax, Vector::new(vec![2.0, 4.0]));
+    /// assert_eq!(cmax, vector![2.0, 4.0]);
     ///
     /// let rmax = a.max(Axes::Row);
-    /// assert_eq!(rmax, Vector::new(vec![3.0, 4.0]));
+    /// assert_eq!(rmax, vector![3.0, 4.0]);
     /// # }
     /// ```
     fn max(&self, axis: Axes) -> Vector<T> where T: Copy + PartialOrd
--- a/src/matrix/base/mod.rs
+++ b/src/matrix/base/mod.rs
@@ -418,6 +418,93 @@ pub trait BaseMatrix<T>: Sized {
                   |sum, row| sum + utils::unrolled_sum(row.raw_slice()))
     }
 
+
+    /// The min of the specified axis of the matrix.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # #[macro_use] extern crate rulinalg; fn main() {
+    /// use rulinalg::matrix::{Matrix, BaseMatrix, Axes};
+    /// use rulinalg::vector::Vector;
+    ///
+    /// let a = matrix![1.0, 2.0;
+    ///                 3.0, 4.0];
+    ///
+    /// let cmin = a.min(Axes::Col);
+    /// assert_eq!(cmin, Vector::new(vec![1.0, 3.0]));
+    ///
+    /// let rmin = a.min(Axes::Row);
+    /// assert_eq!(rmin, Vector::new(vec![1.0, 2.0]));
+    /// # }
+    /// ```
+    fn min(&self, axis: Axes) -> Vector<T> where T: Copy + PartialOrd
+    {
+        match axis {
+            Axes::Col => {
+                let mut mins: Vec<T> = Vec::with_capacity(self.rows());
+                for row in self.row_iter() {
+                    let min = row.iter()
+                                 .skip(0)
+                                 .fold(row[0], |m, &v| if v < m { v } else { m } );
+                    mins.push(min);
+                }
+                Vector::new(mins)
+            },
+            Axes::Row => {
+                let mut mins: Vec<T> = self.row(0).raw_slice().into();
+                for row in self.row_iter().skip(0) {
+                    utils::in_place_vec_bin_op(&mut mins, row.raw_slice(),
+                                              |min, &r| if r < *min { *min = r; });
+                }
+                Vector::new(mins)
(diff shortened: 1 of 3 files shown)


src/matrix/slice.rs
```diff
@@ -414,6 +414,92 @@ pub trait BaseMatrix<T>: Sized {
...
+                for row in self.row_iter() {
+                    let min = row.iter()
+                                 .skip(0)
+                                 .fold(row[0], |m, &v| if v < m { v } else { m } );
+                    mins.push(min);
+                }
+                Vector::new(mins)
+            },
+            Axes::Row => {
+                let mut mins: Vec<T> = self.row(0).iter().map(|&x| x).collect();
```
AtheMathmo: You probably want to use `self.row(0).raw_slice().into()` here - though I'm not sure how much difference it would make.

src/matrix/slice.rs
```diff
@@ -414,6 +414,92 @@ pub trait BaseMatrix<T>: Sized {
...
 
+    /// The min of the specified axis of the matrix.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use rulinalg::matrix::{Matrix, BaseMatrix, Axes};
+    /// use rulinalg::vector::Vector;
+    ///
+    /// let a = Matrix::new(2, 2, vec![1.0, 2.0, 3.0, 4.0]);
```
AtheMathmo: I'd prefer the `matrix!` macro here with each row on it's own line as it makes the dimensions a little easier to see. I mean:

```rust
let a = matrix![1.0, 2.0;
                3.0, 4.0];
```

sinhrks:
Thx, using ``raw_slice`` solved the perf issue.

```
test linalg::matrix::mat_min_col                   ... bench:      15,460 ns/iter (+/- 346)
test linalg::matrix::mat_min_row                   ... bench:      14,076 ns/iter (+/- 253)
```

AtheMathmo:
Awesome!

I'll aim to merge this next week. It will probably be released in 0.4.1 as I'd like to give others some time to check this and I'm ready to release 0.4.

src/matrix/slice.rs
```diff
@@ -181,4 +180,30 @@ mod tests {
         assert_matrix_eq!(a, exp);
 
     }
+
+    #[test]
+    fn matrix_min_max() {
+        let a = matrix!(1., 3., 5., 4.;
```
AtheMathmo: This is minor but can you use square brackets here, please.

We have PR #119 which is switching to square brackets for consistency. I don't want to reduce his hard work before even merging :D.

AtheMathmo:
A minor comment and also a merge conflict here (from the `sum_rows` perf improvement PR I believe).

Once these are resolved I'm happy to merge this.

sinhrks:
Fixed based on the comments, and rebased.

Andlon:
Just a side note on something which should not impact the merge. One unfortunate part of this style of API is that it precludes having a method  `.max()` which merely gives the max value in a matrix slice. This is not a criticism of this PR in particular, but rather the style in which also `mean` and `variance` are written in (since this PR just follows existing conventions).

Somehow I'd prefer a name like `max_along_axis` or similarly `mean_along_axis` to make the functionality a little clearer. Not coming from a predominantly stats/ML background, I would never expect `max` to give the maximum along only an axis, and not for the entire matrix.

Also, one thing to consider for the future would be to perhaps move all these stats-like operations to a trait such as `MatrixStats : BaseMatrix`, which would provide default implementations. Then we could also have them working for matrix slices, and they would also not clutter the `BaseMatrix` trait.

AtheMathmo:
I do agree with your comments @Andlon . I think the API is a little misleading and although I don't like the name `max_along_axis` it might provide a better alternative. Another approach I was considering was to add a new `Axis` variant called `Flat` (or similar). This would have the functionality you suggested. I think this might end up being impractical in some cases though - for example the `split_at` functions.

I also agree with your comment about splitting out the stats functions to a new trait. I worry a little that it makes handling imports for rulinalg more of a headache. Personally I think it is fine but new users may disagree.

I think I will merge the PR as it stands - so that it matches existing functionality more closely. But it seems worthwhile to open an issue to discuss the above. Does that sound ok to you?

Andlon:
@AtheMathmo: Yes, I think that's the right way forward. I think this PR as it stands looks good within the established conventions of the API, and my criticism was mainly for potential future improvements.

I also considered extending the `axis` enum, but I don't think it solves the problem, as Rust does not provide default arguments, and the name `max` (or `mean` for that matter) simply doesn't sound like it would be max along a given axis in the first place. To be clear, I think any function named `max`, `min`, `mean` or similar on a data structure suggests taking the maximum, minimum, mean etc. of all elements in the data structure, and I'd be very surprised to find that this is not the case.

src/matrix/base/mod.rs
```diff
@@ -418,6 +418,92 @@ pub trait BaseMatrix<T>: Sized {
...
+                for row in self.row_iter().skip(0) {
+                    utils::in_place_vec_bin_op(&mut mins, row.raw_slice(),
+                                              |min, &r| if r < *min { *min = r; });
+                }
+                Vector::new(mins)
+            }
+        }
+    }
+
+    /// The min of the specified axis of the matrix.
```
AtheMathmo: Just noticed that this says `min`. I think other than this it is good to merge.

src/matrix/slice.rs
```diff
@@ -181,4 +181,30 @@ mod tests {
         assert_matrix_eq!(a, exp);
 
     }
+
+    #[test]
+    fn matrix_min_max() {
+        let a = matrix![1., 3., 5., 4.;
+                        2., 4., 7., 1.;
+                        1., 1., 0., 0.];
+        assert_eq!(a.min(Axes::Col), Vector::new(vec![1., 1., 0.]));
```
AtheMathmo: Oh - one more. We should probably use the `vector!` macro here.

sinhrks:
oops, fixed typo and used macro in tests.

PR #116, opened 2016-12-25
PERF: Improve perf of matrix.sum_rows

sinhrks:
same impl as #115.

### Before

```
test linalg::matrix::mat_mean_cols                 ... bench:       8,800 ns/iter (+/- 200)
test linalg::matrix::mat_mean_rows                 ... bench:      18,792 ns/iter (+/- 2,265)
test linalg::matrix::mat_sum_cols                  ... bench:       4,915 ns/iter (+/- 1,770)
test linalg::matrix::mat_sum_rows                  ... bench:      12,735 ns/iter (+/- 1,606)
test linalg::matrix::mat_sum_rows_and_cols_128_100 ... bench:      27,507 ns/iter (+/- 9,377)
```

### After 

```
test linalg::matrix::mat_mean_cols                 ... bench:       9,528 ns/iter (+/- 3,713)
test linalg::matrix::mat_mean_rows                 ... bench:      15,885 ns/iter (+/- 4,485)
test linalg::matrix::mat_sum_cols                  ... bench:       4,939 ns/iter (+/- 1,895)
test linalg::matrix::mat_sum_rows                  ... bench:       8,343 ns/iter (+/- 1,591)
test linalg::matrix::mat_sum_rows_and_cols_128_100 ... bench:      21,615 ns/iter (+/- 3,500)
```

### DIFF
--- a/benches/linalg/matrix.rs
+++ b/benches/linalg/matrix.rs
@@ -1,4 +1,4 @@
-use rulinalg::matrix::Matrix;
+use rulinalg::matrix::{Matrix, Axes};
 use rulinalg::matrix::{BaseMatrix, BaseMatrixMut};
 use test::Bencher;
 use test::black_box;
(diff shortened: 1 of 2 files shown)


tafia:
I do not understand why it is faster. Is this some new compiler optimization ?

AtheMathmo:
I also am a little clueless. I know that the compiler can often do some magic with slices - it aids cache access and vectorization. I'm not sure why this particular example gives a speed up though.

I am going to try it on my own architecture to see if I reproduce the results.

AtheMathmo:
Just pinging to let you know there are merge conflicts here. Once resolved I'll check out and see if I can reproduce the performance gain. If yes, will merge this happily.

Do you have any idea why this gives a performance improvement @sinhrks ?

sinhrks:
I **assume** ``row_sums`` is now updated in-place, rather than allocate every-time via ``vec_bin_op``.

AtheMathmo:
Ah that does make sense.

Could you humor me and try using `utils::in_place_vec_bin_op`? So that you update the row sum in place and use explicit vectorization (which in the past has worked better than zip iteration).

sinhrks:
updated to use ``in_place_vec_bin_op``, which can achieve almost the same perf. 

```
test linalg::matrix::mat_mean_cols                 ... bench:       8,814 ns/iter (+/- 3,091)
test linalg::matrix::mat_mean_rows                 ... bench:      14,800 ns/iter (+/- 4,207)
test linalg::matrix::mat_sum_cols                  ... bench:       4,948 ns/iter (+/- 2,381)
test linalg::matrix::mat_sum_rows                  ... bench:       8,419 ns/iter (+/- 997)
test linalg::matrix::mat_sum_rows_and_cols_128_100 ... bench:      22,440 ns/iter (+/- 9,303)
```

AtheMathmo:
I didn't mean that you had to make the change, sorry if that was unclear. It does look like it is more-or-less the same though, which is good. Thanks for checking.

I'm happy to merge this now. How about you @tafia ?

sinhrks:
@AtheMathmo Understood. I intend to use common in-place logic as we might find better in-place impl. I'm going to change #115 to use ``in_place_vec_bin_op`` if OK.

tafia:
Yes of course! Any performance gain is very welcomed. Both the explicit loop and the helper function look good to me.

PR #119, opened 2016-12-26
Unified use of `matrix!` macro (`[…]` vs. `(…)`).

regexident:
While most invocations of the `matrix!` macro in **rulinalg** currently provide parameters in brackets (`[…]`), quite a few of them use parentheses (`(…)`) instead.

I took the liberty to unify them, opting for the more common pattern (also the one matching the common use of `vec![…]`).


src/matrix/impl_ops.rs
```diff
@@ -256,10 +256,10 @@ impl<'a, 'b, T> $trt<&'b T> for &'a Matrix<T>
...
 );
 
-impl_bin_op_scalar_matrix!(Add, add, "addition");
-impl_bin_op_scalar_matrix!(Mul, mul, "multiplication");
-impl_bin_op_scalar_matrix!(Sub, sub, "subtraction");
-impl_bin_op_scalar_matrix!(Div, div, "division");
+impl_bin_op_scalar_matrix![Add, add, "addition"];
+impl_bin_op_scalar_matrix![Mul, mul, "multiplication"];
+impl_bin_op_scalar_matrix![Sub, sub, "subtraction"];
+impl_bin_op_scalar_matrix![Div, div, "division"];
```
Andlon: Maybe these don't actually need to be changed...? I find `(` for function-like macros and `[` for data-like macros to be more fitting, somehow.

regexident: Damnit, that's what one gets for doing find & replace. Fixed.

regexident:
> I haven't checked, but I feel as if there should be a lot more examples of (? I remember when I first wrote the matrix macro I used ( everywhere, but perhaps the other pieces have changed over time.

I just checked, and you were right of course. 😇  Quite sure I got them all now though.

AtheMathmo:
I agree with the above comments. I also prefer `matrix![*]` so thank you for making the change!

Also note that there are some merge conflicts here.

AtheMathmo:
I just had a check and I think a few are still missing - I see some in:

- `matrix/impl_mat`
- `matrix/base/impl_base`
- `matrix/decomposition/eigen`
- `matrix/decomposition/svd`

Am happy to update these myself if you don't have time.

regexident:
> I just had a check and I think a few are still missing.

Hm… had to do major conflict resolving against 7d6b58bc729e086f5fd28b5636dd093429977768. Must have missed them. Fixed now.

AtheMathmo:
Fixed the merge conflict. Will merge this in once CI approves. Thank you!

PR #120, opened 2016-12-26
Added `vector![…]` macro mirroring `matrix![…]`

regexident:
We have a `matrix!` macro for ergonomic matrix declarations. Vector should have the same, imho.

### DIFF
--- a/src/macros/vector.rs
+++ b/src/macros/vector.rs
@@ -1,0 +1,100 @@
+/// The `vector!` macro enables easy construction of small vectors.
+///
+/// This is particularly useful when writing tests involving vectors.
+/// Note that the macro is just a convenient wrapper around the Vector
+/// constructors, and as a result the vector is still allocated on the
+/// heap.
+///
+/// # Examples
+///
+/// ```
+/// #[macro_use]
+/// extern crate rulinalg;
+///
+/// # fn main() {
+/// use rulinalg::vector::Vector;
+///
+/// // Construct a vector of f64
+/// let vec = vector![1.0, 2.0, 3.0];
+/// # }
+/// ```
+///
+/// To construct vectors of other types, specify the type by
+/// the usual Rust syntax:
+///
+/// ```
+/// #[macro_use]
+/// extern crate rulinalg;
+///
+/// # fn main() {
+/// use rulinalg::vector::Vector;
+///
+/// // Construct a vector of f32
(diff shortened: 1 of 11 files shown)


tests/mat/mod.rs
```diff
@@ -39,7 +39,7 @@ fn test_u_triangular_solve_errs() {
     assert!(a.solve_u_triangular(Vector::new(vec![])).is_err());
```
Andlon: Can also replace this one with `vector![]`, I think?

Andlon:
Yes, this is great! I've been wanting to implement a `vector!` macro for a long time, but never got around to it.

tests/mat/mod.rs
regexident: Oh, I missed that one (and line 30). Fixed.

AtheMathmo:
Thanks! This is a very welcome addition.

Just to note - there are some merge conflicts here currently.

src/macros/vector.rs
```diff
@@ -0,0 +1,88 @@
...
+/// the usual Rust syntax:
+///
+/// ```
+/// # #[macro_use]
+/// # extern crate rulinalg;
+/// # fn main() {
+/// use rulinalg::vector::Vector;
+///
+/// // Construct a vector of f32
+/// let mat: Vector<f32> = vector![1.0, 2.0, 3.0];
```
AtheMathmo: I think we should call this `vector` not `mat`. Or just `v`?

src/macros/vector.rs
```diff
@@ -0,0 +1,88 @@
...
+/// // Or
+/// let mat = vector![1.0, 2.0, 3.0f32];
```
AtheMathmo: And here.

src/macros/vector.rs
```diff
@@ -0,0 +1,88 @@
...
+    ($x:expr; $n:expr) => {
+        {
+            use $crate::vector::Vector;
+            Vector::new(vec![$x; $n])
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
AtheMathmo: Just for the sake of completeness we should also test the constant size expansion: `vector![1.0; 5]`.

AtheMathmo:
This looks ready to merge now to me. Thank you!

I'll give @Andlon a few hours in case he has any more comments. If not I'll merge this in.

Andlon:
Looks good to me! :+1:

Issue #110, opened 2016-12-17
Moore–Penrose pseudo-inverse of Matrix


regexident:
I was trying to port some code to Rust that required computing the pseudo-inverse of a matrix and had to realize that **rulinalg** currently does not provide such functionality.

According to [Wikipedia](https://en.wikipedia.org/wiki/Singular_value_decomposition#Applications_of_the_SVD) the pseudo-inverse of a matrix can be implemented using singular value decomposition (SVD), which is already available in **rulinalg**.

I attempted an implementation myself by porting some Coffeescript impl (for complex matrices):

```coffeescript
cpinv = (A) ->
    {U, S, V} = csvd A
    tol = max(size(A))*nm.epsilon*S[0]
    Z = ((if x>tol then 1/x else 0) for x in S)
    V.dot(nm.diag(Z)).dot(U.transjugate())
```

which I found [on Github](https://github.com/stemblab/complex-pseudoinverse/blob/gh-pages/main.coffee#L11-L15) to **rulinalg** (stripping it down to work with plain old real values), but got stuck at the use of `dot` in the last line:

```rust
impl<T: Any + Float + Signed + FromPrimitive> Matrix<T> {
    pub fn pseudo_inverse(self) -> Result<Matrix<T>, Error> {
        self.svd().map(|(s, u, v)| {
            let max = T::from_usize(::std::cmp::max(self.cols(), self.rows())).unwrap();
            let s_zero = s.get_unchecked([0, 0]).clone();
            let epsilon = max * T::epsilon() * s_zero;
            let z = s.diag().apply(&|x| if x > epsilon { T::one() / x } else { T::zero() });
            v.dot(z).dot(u.transpose())
        })
    }
}
```

Any idea how the use of `dot` would translate to **rulinalg** (or if I'm on the right track to begin with)?

AtheMathmo:
I haven't done any background reading into the topic here-  but I imagine that the `dot` function translates to a matrix multiplication. For this you can just use the `*` operator:

```rust
v * z * u.transpose()
```

I'll try to read over the topic and code properly at some point if that doesn't help!

regexident:
Thanks for the prompt response, AtheMathmo!

Yeah, that's what I came up with, too. What tripped me up with though was the `z`-vector:

```rust
(v: Matrix) * (z: Vector) * (u.transpose(): Matrix)
```

There is no `vector * matrix` that I know of.

---

[...]

PR #122, opened 2016-12-26
[WIP] Implementation of Moore–Penrose pseudo-inverse

regexident:
Aims to implement issue https://github.com/AtheMathmo/rulinalg/issues/110.

---

Feedback appreciated.

1. Doesn't work with tall matrices, apparently.
2. Might be a good idea to fall back to faster plain inverse for square matrices?

---

More info on [Moore–Penrose pseudo-inverse](http://math.stackexchange.com/questions/624196/calculating-moore-penrose-pseudo-inverse).

### DIFF
--- a/src/matrix/impl_mat.rs
+++ b/src/matrix/impl_mat.rs
@@ -1,10 +1,10 @@
 use std::any::Any;
 use std::fmt;
-use libnum::{One, Zero, Float, FromPrimitive};
+use libnum::{One, Zero, Float, FromPrimitive, Signed};
 
 use super::{Matrix, forward_substitution, back_substitution, parity};
 use super::{Axes};
-use super::base::BaseMatrix;
+use super::base::{BaseMatrix, BaseMatrixMut};
 use error::{Error, ErrorKind};
 use vector::Vector;
 
@@ -18,7 +18,8 @@ impl<T> Matrix<T> {
     /// ```
     /// use rulinalg::matrix::{Matrix, BaseMatrix};
     ///
-    /// let mat = Matrix::new(2,2, vec![1.0,2.0,3.0,4.0]);
+    /// let mat = Matrix::new(2,2, vec![1.0, 2.0,
+    ///                                 3.0, 4.0]);
     ///
     /// assert_eq!(mat.rows(), 2);
     /// assert_eq!(mat.cols(), 2);
@@ -314,7 +315,8 @@ impl<T: Any + Float> Matrix<T> {
     /// use rulinalg::matrix::Matrix;
     /// use rulinalg::vector::Vector;
     ///
-    /// let a = Matrix::new(2,2, vec![2.0,3.0,1.0,2.0]);
+    /// let a = Matrix::new(2,2, vec![2.0, 3.0,
+    ///                               1.0, 2.0]);
     /// let y = Vector::new(vec![13.0,8.0]);
     ///
     /// let x = a.solve(y).unwrap();
@@ -343,14 +345,16 @@ impl<T: Any + Float> Matrix<T> {
     /// # Examples
     ///
     /// ```
+    /// # #[macro_use] extern crate rulinalg; fn main() {
     /// use rulinalg::matrix::Matrix;
     ///
-    /// let a = Matrix::new(2,2, vec![2.,3.,1.,2.]);
+    /// let a = Matrix::new(2, 2, vec![2.0, 3.0,
+    ///                                1.0, 2.0]);
     /// let inv = a.clone().inverse().expect("This matrix should have an inverse!");
-    ///
-    /// let I = a * inv;
-    ///
-    /// assert_eq!(*I.data(), vec![1.0,0.0,0.0,1.0]);
+    /// let expected = Matrix::new(2,2, vec![2.0, -3.0;
+    ///                                     -1.0,  2.0]);
+    /// assert_matrix_eq!(inv, expected);
+    /// # }
     /// ```
     ///
     /// # Panics
@@ -408,8 +412,8 @@ impl<T: Any + Float> Matrix<T> {
     /// ```
     /// use rulinalg::matrix::Matrix;
     ///
-    /// let a = Matrix::new(3,3, vec![1.0,2.0,0.0,
-    ///                               0.0,3.0,4.0,
+    /// let a = Matrix::new(3,3, vec![1.0, 2.0, 0.0,
+    ///                               0.0, 3.0, 4.0,
     ///                               5.0, 1.0, 2.0]);
     ///
     /// let det = a.det();
@@ -468,6 +472,35 @@ impl<T: Any + Float> Matrix<T> {
             sgn * d
         }
     }
+}
+
+impl<T: Any + Float + Signed + FromPrimitive> Matrix<T> {
+    /// Computes the pseudo-inverse of the matrix.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # #[macro_use] extern crate rulinalg; fn main() {
+    /// use rulinalg::matrix::Matrix;
+    ///
+    /// let a: Matrix<f32> = Matrix::new(2, 2, vec![2.0, 3.0,
+    ///                                             1.0, 2.0]);
+    /// let pinv = a.pseudo_inverse().expect("This matrix should have a pseudo-inverse!");
+    /// let expected = Matrix::new(2, 2, vec![2.0, -3.0,
+    ///                                      -1.0,  2.0]);
+    /// assert_matrix_eq!(pinv, expected, comp = float);
+    /// # }
+    /// ```
+	pub fn pseudo_inverse(self) -> Result<Matrix<T>, Error> {
+		let max_side = ::std::cmp::max(self.cols(), self.rows());
+		self.svd().map(|(s, u, v)| {
+			let max = T::from_usize(max_side).unwrap();
+			let s_max = unsafe { s.get_unchecked([0, 0]).clone() };
+			let epsilon = max * T::epsilon() * s_max;
+			let z = s.apply(&|x| if x > epsilon { T::one() / x } else { T::zero() });
+			v * z * u.transpose()
+		})
+	}
 }
 
 impl<T: fmt::Display> fmt::Display for Matrix<T> {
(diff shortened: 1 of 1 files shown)


src/matrix/mod.rs
```diff
@@ -734,6 +734,35 @@ impl<T: Any + Float> Matrix<T> {
...
+    /// ```
+    /// use rulinalg::matrix::Matrix;
+    ///
+    /// let a: Matrix<f32> = Matrix::new(2, 2, vec![2.0, 3.0, 1.0, 2.0]);
+    /// let pinv = a.clone().pseudo_inverse().expect("This matrix should have a pseudo-inverse!");
+    ///
+    /// let i = a * pinv;
+    ///
+    /// for (a, b) in (*i.data()).iter().zip(vec![1.0, 0.0, 0.0, 1.0]) {
+    ///     assert!((*a - b).abs() < 0.0001);
```
Andlon: You can use the [assert_matrix_eq!](https://docs.rs/rulinalg/0.3.7/rulinalg/macro.assert_matrix_eq.html) macro to clean up the assertion here.

src/matrix/mod.rs
```diff
@@ -734,6 +734,35 @@ impl<T: Any + Float> Matrix<T> {
...
+    /// }
+    /// ```
+    pub fn pseudo_inverse(self) -> Result<Matrix<T>, Error> {
+        let max_side = ::std::cmp::max(self.cols(), self.rows());
+        self.svd().map(|(s, u, v)| {
+            let max = T::from_usize(max_side).unwrap();
+            let s_max = unsafe { s.get_unchecked([0, 0]).clone() };
```
Andlon: Is it perhaps possible that `s`, `u` and `v` are all empty matrices (if `self` is empty), in which case this unsafe block is actually not safe? I'm not sure if the SVD currently works this way, but if it doesn't, it should!

src/matrix/mod.rs
```diff
@@ -1122,6 +1151,14 @@ mod tests {
     }
 
     #[test]
+    fn matrix_inverse() {
+        let a = Matrix::new(2, 2, vec![2., 3., 1., 2.]);
+        let inv = a.inverse().expect("This matrix should have an inverse!");
+        let i = a * inv;
+        assert_eq!(*i.data(), vec![1.0, 0.0, 0.0, 1.0]);
```
Andlon: Can also use `assert_matrix_eq` here

src/matrix/mod.rs
```diff
@@ -1157,6 +1194,21 @@ mod tests {
...
+        let a: Matrix<f32> = matrix!(
+            1.0, 2.0, 3.0;
+            4.0, 5.0, 6.0
+        );
+        let b = a.clone().pseudo_inverse().expect("This matrix should have a pseudo-inverse!");
+        assert_eq!(a.cols(), b.rows());
+        assert_eq!(a.rows(), b.cols());
+        let c = a * b;
+        for (x, y) in c.iter().zip(matrix!(1.0, 0.0; 0.0, 1.0).iter()) {
+            assert!((x - y).abs() < 0.0001);
```
Andlon: `assert_matrix_eq` here too, and please test against expected pseudo-inverse directly, rather than the product.

src/matrix/mod.rs
```diff
@@ -734,6 +734,35 @@ impl<T: Any + Float> Matrix<T> {
...
+    /// for (a, b) in (*i.data()).iter().zip(vec![1.0, 0.0, 0.0, 1.0]) {
+    ///     assert!((*a - b).abs() < 0.0001);
+    /// }
+    /// ```
+    pub fn pseudo_inverse(self) -> Result<Matrix<T>, Error> {
+        let max_side = ::std::cmp::max(self.cols(), self.rows());
+        self.svd().map(|(s, u, v)| {
+            let max = T::from_usize(max_side).unwrap();
+            let s_max = unsafe { s.get_unchecked([0, 0]).clone() };
+            let epsilon = max * T::epsilon() * s_max;
```
Andlon: I can see why you multiply by `s_max`, but I don't follow why you multiply by `max`. Can you elaborate? Perhaps preferably in the main PR comment section, so that it doesn't get lost in review comments!

regexident:
> I also don't quite understand your epsilon criterion, as I've noted in one of my review comments. Could you perhaps elaborate on why you multiply by the maximum row/col dimension as well? I'm curious!

Allow me to quote from Quora:

> Sometimes we would like to invert a matrix which is not quite invertible. A useful alternative is to invert the part of the matrix that is invertible (its image) and to leave alone the part of the matrix that is not invertible (its kernel), by applying a real valued function `if x > epsilon { 1 / x } else { 0 }` to a symmetric matrix , the resulting matrix `A⁺` is called pseudo inverse.for an invertible matrix `A⁺=A⁻¹`. ([Source](https://www.quora.com/What-is-the-intuition-behind-pseudo-inverse-of-a-matrix/answer/Sameer-Gupta))

I am by no means an expert in linalg. I basically looked for existing implementations based on SVD and then implemented a version using rulinalg's API.

> I'm also a little curious when one would actually want to use the pseudo inverse. I've always considered it primarily a theoretical tool. There are often much more efficient ways to accomplish the same as any of its applications, but I suppose it could be really useful for quick prototyping?

My need for `fn pseudo_inverse`comes from writing a Rust version of the [Smoothed L0 algorithm](http://ee.sharif.edu/~SLzero/).

AtheMathmo:
> svd should also never fail. @AtheMathmo: could you perhaps give your thoughts on this?

Indeed it should never fail but I believe the current algorithm has some issues which may cause it to fail in edge-cases. Some reasons why:

- We cannot reduce the matrix to bidiagonal form (presumably because the reduction algorithm is also lacking).
- We fail to compute eigenvalues at one step (again because our reduction algorithms fail OR there is an implementation bug).

It may be better to have the library panic if one of these fails. I decided at the time it was best to note the issues in the documentation and let the user decide how to handle these algorithmic failings.

Andlon:
Thanks for the references, @regexident! I still don't understand why one multiplies by `max(rows, cols)` in the `epsilon` criterion, however. It seems to me that this can lead to some really unfortunate corner cases. Consider for example a matrix of size `1000000 x 10`, then the epsilon is multiplied by `1e6`, and this doesn't seem to be balanced by any other factor, so you'd just end up with a much looser tolerance... Still, MATLAB uses this, so I'd assume it is for a reason. I just wish I could see why!

Also, thanks for providing me with your motivation of what you want to use the pseudo inverse for!

Andlon:
@AtheMathmo: thanks for providing some insight. As you know, I don't have much time to work on rulinalg these days, otherwise I'd like to take a crack at improving the SVD implementation. I've picked up some relevant experience lately when working on my thesis.

AtheMathmo:
Well if you do get the time I would really appreciate the impact of more experienced hands!

I'd also like to go back and try to improve things there, especially as SVD is such an important algorithm. Sadly I'm not sure I can do much more without a lot of extra reading (which I also do not have time for right now).

regexident:
Huh, I didn't mean to close this?

src/matrix/impl_mat.rs
```diff
@@ -470,6 +470,35 @@ impl<T: Any + Float> Matrix<T> {
...
+	///
+	/// ```
+	/// use rulinalg::matrix::Matrix;
+	///
+	/// let a: Matrix<f32> = Matrix::new(2, 2, vec![2.0, 3.0, 1.0, 2.0]);
+	/// let pinv = a.clone().pseudo_inverse().expect("This matrix should have a pseudo-inverse!");
+	///
+	/// let i = a * pinv;
+	///
+	/// for (a, b) in (*i.data()).iter().zip(vec![1.0, 0.0, 0.0, 1.0]) {
```
AtheMathmo: I'd also prefer to see the `assert_matrix_eq!` macro here.

src/matrix/impl_mat.rs
```diff
@@ -470,6 +470,33 @@ impl<T: Any + Float> Matrix<T> {
     }
 }
 
+impl<T: Any + Float + Signed + FromPrimitive> Matrix<T> {
+	/// Computes the pseudo-inverse of the matrix.
+	///
+	/// # Examples
+	///
+	/// ```
+    /// # #[macro_use] extern crate rulinalg; fn main() {
```
AtheMathmo: I think there are some indentation inconsistencies in this example.

src/matrix/impl_mat.rs
```diff
@@ -470,6 +470,33 @@ impl<T: Any + Float> Matrix<T> {
...
+	/// use rulinalg::matrix::Matrix;
+	///
+	/// let a: Matrix<f32> = Matrix::new(2, 2, vec![2.0, 3.0, 1.0, 2.0]);
```
AtheMathmo: Can we switch to the `matrix!` macro here, please.

src/matrix/impl_mat.rs
```diff
@@ -343,14 +343,14 @@ impl<T: Any + Float> Matrix<T> {
     /// # Examples
     ///
     /// ```
+    /// # #[macro_use] extern crate rulinalg; fn main() {
     /// use rulinalg::matrix::Matrix;
     ///
-    /// let a = Matrix::new(2,2, vec![2.,3.,1.,2.]);
+    /// let a = matrix![2., 3.; 1., 2.];
```
AtheMathmo: I would prefer to see this matrix spread over 2 lines:
```rust
matrix![2., 3.;
        1., 2.];
```

Andlon:
@AtheMathmo: I think a problem here is that the implementation cannot be made to work correctly because the SVD implementation doesn't work correctly. Merging now means we'll add one more semi-broken implementation to the library, which might also have to be fixed once the SVD implementation has been fixed... I'm not sure how best to resolve this.

AtheMathmo:
You have a good point...

It gets increasingly difficult to maintain the library if we add more code that comes with disclaimers.

AtheMathmo:
I am closing this due to the issues discussed above. If we see SVD improvements then this PR will probably be very easy to fix up and merge.

I will keep #110 open even so.

regexident:
While sad I totally understand the reasoning for this decision. 
Feel free to drop me a line once things have settled/been resolved and I'll be happy to create another PR. 🙂

Andlon:
@regexident: It's certainly on my TODO list to revisit the SVD implementation and hopefully fix it (unless anyone else can take a crack at it first!). However, I'm in the process of revamping the API for decompositions, and I will probably save SVD for last (because it's by far the most challenging/complicated one).

AtheMathmo:
To be clear - I'd also very much like to see this PR merged in the future. But right now I need to slim down the mental load of all the PRs (on this project and others).

Hopefully @Andlon can crack the SVD issues and we can take advantage here and elsewhere!

regexident:
Sounds good to me! Maybe give the PR a tag like "revisit" or something?

regexident:
FYI, I just stumbled upon another application of the Moore-Penrose pseudo-inverse:
It is used in the "[Jacobian inverse technique](https://en.wikipedia.org/wiki/Inverse_kinematics#The_Jacobian_inverse_technique)" as used in Inverse Kinematics (IK).

PR #123, opened 2016-12-27
ENH/CLN: Add Vector ops

sinhrks:
cleaned up vector ops using macro, adding some missings (`Rem`, `Not`, and `BitXXX`).

The existing behaviors are basically the same, but I removed the ``div`` assertion whether divider is 0 (as ``Matrix`` currently doesn't check it). 
- https://github.com/AtheMathmo/rulinalg/blob/master/src/vector.rs#L536


AtheMathmo:
Thank you for the cleanup! It has been needed for a while.

I've had a brief glance through but will do a proper review soon. I just have one minor point. Recently I changed the `matrix/mod` file to only contain definitions and moved the `impl Matrix` blocks to a new `impl_mat` module. I think this is less important for `Vector` but unless you see a reason to avoid it we might prefer to be consistent.

sinhrks:
It sounds better. Split ``impl_vec`` to impls.

src/vector/impl_ops.rs
```diff
@@ -0,0 +1,1374 @@
+use std::ops::{Mul, Add, Div, Sub, Rem,
+               MulAssign, AddAssign, DivAssign, SubAssign, RemAssign,
+               Neg, Not,
+               BitAnd, BitOr, BitXor, BitAndAssign, BitOrAssign, BitXorAssign};
+use utils;
+
+use super::Vector;
+
+impl<T: Copy + Mul<T, Output = T>> Vector<T> {
```
AtheMathmo: In my opinion we should keep `elemul` and `elediv` in the `impl_vec` module.

I agree that they are operations but as Rust doesn't directly support them it will be easier to keep track of them in the other module. For example if we ever make a `BaseVector` slice we'd like to include these functions.

sinhrks: OK, moved them and corresponding tests to ``impl_vec``

AtheMathmo:
I'll review properly later today or tomorrow. It looks good at a glance though.

Thanks for this!

src/vector/impl_vec.rs
```diff
@@ -465,305 +452,6 @@ impl<T: Float + FromPrimitive> Vector<T> {
...
-
-    fn neg(self) -> Vector<T> {
-        let new_data = self.data.iter().map(|v| -*v).collect::<Vec<_>>();
-
-        Vector::new(new_data)
-    }
-}
-
 /// Indexes vector.
 impl<T> Index<usize> for Vector<T> {
```
AtheMathmo: Did you intend to leave this here?

I'm not too committed to placing it in either file.

sinhrks: Yeah, but it looks better to move it to ``impl_ops`` to be consistent with ``Matrix``. Done.

AtheMathmo:
I'm afraid I might have given you quite a nasty merge conflict to deal with. We had two vector PRs which changed quite a lot. I believe the conflict is mostly removing the `Vector::new(vec![...])` construct in place of the new `vector![...]` macro.

If you want me to handle the conflict let me know, I'm happy to do so.

sinhrks:
NP:) Updated based on the latest master.

AtheMathmo:
Looks good to merge to me.

PR #125, opened 2016-12-28
API: Row/Column to Vector conversion

sinhrks:
Provides easier conversion from ``Row/Column`` to ``Vector``.


src/matrix/impl_conv.rs
```diff
@@ -0,0 +1,100 @@
...
+);
+impl_row_to_vector!(Row);
+impl_row_to_vector!(RowMut);
+
+macro_rules! impl_column_to_vector (
+    ($t:ident) => (
+
+impl<'a, T: Clone> From<$t<'a, T>> for Vector<T> {
+    fn from(column: $t<'a, T>) -> Vector<T> {
+        column.iter().map(|x| x.clone()).collect::<Vector<T>>()
```
Andlon: Super minor and not at all important, but couldn't you perhaps just use `iter().cloned().collect`?

AtheMathmo:
Is there a reason you chose not to use the `convert` module? To me it seems a good idea to keep these in the same place - sorry if I'm missing something.

Otherwise this looks good to me and is a welcome addition! You should also make the change @Andlon suggested - just as it looks a little prettier :).

sinhrks:
I haven't noticed ``convert.rs`` :) Moved the impl fixing to use ``cloned``.

AtheMathmo:
This looks good to merge to me.

AtheMathmo:
I'll merge this now.

I agree with your comments @Andlon - these would all be useful additions. We could implement these without too much fuss now but the whole story would be much nicer if we had a few new rust features, like specialization and custom dynamically sized pointers.

PR #126, opened 2016-12-28
ENH: Add Matrix CSV reader / writer

sinhrks:
This PR allows to read / write matrix using CSV file. 

I'd like to discuss what the API should be. I think  ``read`` and ``write`` should be methods of ``Matrix``. We can use ``Matrix::<f64>::read(reader) `` or ``mat.write(writer)``, and naturally we can support other types than ``f64``.


Andlon:
Pretty cool!

As far as I know, IO hasn't really been discussed before. Since it's a pretty big topic in terms of API, perhaps an issue for the discussion of a preliminary IO API should be made, referring to this PR as a baseline for discussion?

Cargo.toml
```diff
@@ -10,9 +10,13 @@ keywords = ["linalg","linear","algebra","linear-algebra","matrix"]
 readme = "README.md"
 license = "MIT"
 
+[features]
+io = []
+
 [dependencies]
 num = {version = "0.1.36", default-features = false }
 matrixmultiply = "0.1.13"
+csv = "0.14.7"
```
AtheMathmo: I wonder how this dependency will work against the feature. Will it only be downloaded when the `io` feature is given? If not, it would be heavily preferred if we could support this.

src/io/io_csv.rs
```diff
@@ -0,0 +1,112 @@
...
+pub fn read<'a, R: Read>(mut reader: Reader<R>) -> Matrix<f64> {
+    // headers read 1st row regardless of has_headers property
+    let header : Vec<String> = reader.headers().unwrap();
+
+    let mut nrows = 0;
+    let ncols = header.len();
+
+    let mut records: Vec<f64> = Vec::with_capacity(header.len() * 100);
+    for record in reader.decode() {
+        let values: Vec<f64> = record.unwrap();
```
AtheMathmo: We probably want to handle the case in which this panics. I think it would be better to have `read` return a `Result`. At the very least we would prefer a custom error message.

src/io/io_csv.rs
```diff
@@ -0,0 +1,112 @@
...
+        records.extend(values);
+        nrows += 1;
+    }
+    records.shrink_to_fit();
+    Matrix::new(nrows, ncols, records)
```
AtheMathmo: Is it possible for any one row to have more elements? I would imagine so.

In which case we _could_ get a panic here which would give a non-obvious message to our user. Perhaps we should add an assertion in the inner loop to ensure that the row contains the expected number of elements, and fail (with a panic or `Err` return) otherwise.

src/io/io_csv.rs
```diff
@@ -0,0 +1,112 @@
...
+}
+
+/// Write Matrix<f64> as csv file
+pub fn write<'a, W: Write>(writer: &'a mut Writer<W>, inputs: &Matrix<f64>) {
```
AtheMathmo: I wonder if we should have this function work with a `BaseMatrix`? The code would remain the same.

I think it would be a rare use-case but it is no additional work for us.

src/io/mod.rs
```diff
@@ -0,0 +1 @@
+pub mod io_csv;
```
AtheMathmo: I think this name is a little unconventional. I think I'd prefer `io::csv` to `io::io_csv`.

Any other opinions?

AtheMathmo:
Oh and one more thing - we should document that this module handles float-only csv files. I think this is a sensible choice to make but people may expect to be able to do more advanced things like category-based data columns if we don't make it clear.

src/io/io_csv.rs
```diff
@@ -0,0 +1,112 @@
...
+
+/// Read csv file as Matrix<f64>
+pub fn read<'a, R: Read>(mut reader: Reader<R>) -> Matrix<f64> {
+    // headers read 1st row regardless of has_headers property
+    let header : Vec<String> = reader.headers().unwrap();
+
+    let mut nrows = 0;
+    let ncols = header.len();
+
+    let mut records: Vec<f64> = Vec::with_capacity(header.len() * 100);
```
Andlon: I think in general, if one doesn't know an upper bound for the size of a vector, it's better to avoid specifying a capacity at all.

src/io/io_csv.rs
```diff
@@ -0,0 +1,112 @@
+//! csv read / write module
+use csv::{Reader, Writer};
+use std::io::{Read, Write};
+
+use super::super::matrix::{Matrix, BaseMatrix};
+
+/// Read csv file as Matrix<f64>
+pub fn read<'a, R: Read>(mut reader: Reader<R>) -> Matrix<f64> {
```
Andlon: Is it perhaps possible to take `f64` as a type parameter, letting the user parse into the desired type...?

src/io/io_csv.rs
```diff
@@ -1,39 +1,67 @@
...
-        records.extend(values);
-        nrows += 1;
+impl<T> Matrix<T> where T: Decodable {
+
+    /// Read csv file as Matrix
+    pub fn read_csv<'a, R: Read>(mut reader: csv::Reader<R>)
+        -> Result<Matrix<T>, csv::Error> {
+
+        // headers read 1st row regardless of has_headers property
+        let header: Vec<String> = match reader.headers() {
```
AtheMathmo: These match blocks can be written as `try!` statements.

src/io/mod.rs
sinhrks: Because end user use the func with ``csv`` crate, using the same name looks confusing.

sinhrks:
Updated a bit. Now ``Matrix`` has ``read_csv`` and ``write_csv`` which accepts ``Reader/Writer``. And we can specify input type like ``Matrix::<f64>::read_csv``.

src/io/io_csv.rs
```diff
@@ -0,0 +1,159 @@
...
+}
+
+impl<T> Matrix<T> where T: Encodable {
+
+    /// Write Matrix<f64> as csv file
+    pub fn write_csv<W: Write>(&self, writer: &mut csv::Writer<W>)
+        -> Result<(), csv::Error> {
+
+        for row in self.row_iter() {
+            match writer.encode(row.raw_slice()) {
```
AtheMathmo: And we can use `try!` here too.

AtheMathmo:
I think that this looks pretty good now.

I think we should open up an issue to try and get people talking about what they'd expect and whether this PR is missing anything or doing anything wrong. Personally I am pretty happy with what is presented here, but I'm sure there are things to discuss like serialization, etc.

AtheMathmo:
@Andlon would you mind taking one last look at this before I merge?

The only thing I've noticed is that we may want to have the `travis.yml` file run only the `io` feature version (instead of 2 runs - the second will run all of the first tests anyway).

src/io/mod.rs
```diff
@@ -0,0 +1 @@
+mod io_csv;
```
Andlon: I still feel the `io_` prefix is a little redundant here? Since its parent is already `io`.

src/io/io_csv.rs
```diff
@@ -0,0 +1,181 @@
...
+        wtr.flush().unwrap();
+
+        let rdr = csv::Reader::from_file("./data.csv").unwrap().has_headers(false);
+        let res = Matrix::<f64>::read_csv(rdr).unwrap();
+        assert_matrix_eq!(mat, res);
+
+        // remove file;
+        use std::fs;
+        fs::remove_file("./data.csv").unwrap();
+    }
```
Andlon: File I/O tests are a bit tricky, because they involve state outside of the test binary. In particular, this test will write a file `data.csv` in the current directory, and delete it afterwards. This is problematic for the following reasons:

- The current directory could be almost anything depending on how the tests are invoked, and so executing tests can place a file in an almost arbitrary location on the file system, which is not very nice for the unsuspecting user.
- If anything fails between creating/writing to the file and removing the fail, the file will not be removed. For example, if the assertion `assert_matrix_eq!(mat, res)` fails.
- Each test is not necessarily independent of previous runs. In the above test, there could for example be a problem with the file system which somehow makes writing fail silently, so that you end up reading old data from a previous test. Granted, this is unlikely, but it makes the test more brittle.

Now, to remedy this, you would need some kind of temporary file utility. I haven't tried it, but [tempfile](https://stebalien.github.io/tempfile/tempfile/) looks pretty good. The idea is then to construct a temporary file, and with the help of the destructor of the tempfile utility, the file will be deleted even if a panic occurs in the test.

Finally, we _might_ want to move this into the `tests` folder since it interfaces with the file system. It makes no practical difference here, but in general one tends to classify tests which interface with external resources as integration tests.

src/io/mod.rs
AtheMathmo: Ah yes, I forgot about this. I also had this issue before and I do think we should change it. Although repeating csv may be confusing for the end user as they are interfacing with the csv crate I think it is more important to follow naming conventions here.

AtheMathmo: Perhaps one reasonable solution is to reexport the csv crate readers and writers in our module?

Andlon: Since we're already tied to the `csv` API, I think that could be a good idea, yeah! Then users (almost) wouldn't even really need to care about the fact that `csv` is an external crate at all.

src/io/io_csv.rs
AtheMathmo: I agree with these concerns too. I think in fact we could argue that this test isn't really testing any rulinalg functionality at all. The other tests cover rulinalg behaviour for different data and here it seems we are testing that the csv readers behave consistently when using file system io.

Because of the potential issues discussed above I am ok with seeing this test completely removed. Otherwise we can try to make it safer using some of @Andlon 's suggestions above.

Andlon: Agree with your assessment!

sinhrks:
OK, updated to re-export ``Reader/Writer`` form ``rulinalg::io::csv::``.

src/lib.rs
```diff
@@ -110,3 +110,12 @@ extern crate itertools;
 
 pub use norm::{VectorNorm, MatrixNorm};
 pub use norm::{VectorMetric, MatrixMetric};
+
+#[cfg(feature = "io")]
+extern crate csv as mod_csv;
```
AtheMathmo: This rename doesn't make sense to me. `mod_csv` seems to suggest this is a module within rulinalg. I would instead name is `libcsv`?

tests/mat/csv.rs
```diff
@@ -0,0 +1,96 @@
+use rulinalg::matrix::Matrix;
+use rulinalg::io::csv::{Reader, Writer};
+
+#[test]
+fn test_read_csv_with_header() {
```
AtheMathmo: I think there is some confusion around these tests.

I believe that @Andlon was only suggesting we move the file system test into the integration tests folder. The other tests are fine in the `io::csv` module. As we have deleted the file system test I think it is safe to leave these tests where they were before.

Is this what you meant @Andlon ?

sinhrks: i intend it is better to test here using the api exposed to users.

src/lib.rs
sinhrks: i have no preference. will change.

tests/mat/csv.rs
Andlon: @AtheMathmo: yes, that's what I meant, but I think @sinhrks makes a good point. Because the `csv` module integrates its API with an external crate, it's useful to test it from *outside* the rulinalg crate itself (which the tests in the `tests` folder do), because this more closely mirrors how it's used by users.

So I actually agree that it's a good idea to put these tests here, although perhaps they should be in a new `io` subfolder rather than `mat`?

Andlon: (things seem to be a little out of sync with Github on my end?)

@AtheMathmo: yes, that's what I meant, but I think @sinhrks makes a good point. Because the `csv` module integrates its API with an external crate, it's useful to test it from *outside* the rulinalg crate itself (which the tests in the `tests` folder do), because this more closely mirrors how it's used by users.

So I actually agree that it's a good idea to put these tests here, although perhaps they should be in a new io subfolder rather than mat?

AtheMathmo: I think all of our nitpicking has been resolved :).

Let me know if you spot anything else. If not I'll be happy to get this in!

Andlon:
Looks good to me too. Thanks for implementing this useful feature, @sinhrks!

PR #133, opened 2016-12-29
Ignore Eclipse ".project" file

c410-f3r:
Eclipse automatically creates a file named ".project" for every single project, just like Visual Studio Code creates a ".vscode" file for workspace settings.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ Cargo.lock
 .vscode
 *.rs.bk
 .DS_Store
+.project


AtheMathmo:
I am more than happy to merge this - but we've had a few of these now and I wonder if we can improve the situation.

Is there any reason why we cannot simply add `.*` to the gitignore file to cover all such files? We _do_ have the `.travis.yml` file which may need updates - can we handle this still?

andrewcsmith:
Just to chime in – I don't really appreciate extra content related to particular IDEs that I don't have maintained in repos I clone.

I'd point people to [Explicit repository excludes](https://help.github.com/articles/ignoring-files/#explicit-repository-excludes).

AtheMathmo:
I think there is a balance to be struck. For example, so many people use vim that it makes sense to have this handled in the gitignore file.

Personally I don't find it too problematic to merge in one of these PRs every now and then :)

c410-f3r:
I think the `.gitignore` approach is harder to maintain but feels more practical. Anyway, I am cool with any convention that is going to be established.

AtheMathmo:
I'm merging this for now. If this arises again then I will add a note to the contributing docs and advise others to use the repository excludes as suggested by @andrewcsmith .

Thank you both.

PR #137, opened 2017-01-02
Add column iterator

c410-f3r:
I still don't know for sure what the `*_stride` attribute does but the column iterator is working as expected without it.
Didn't include `raw_slice` or similar for `Column` and `ColumnMut` since it is not possible to return slices for non-contiguous data. We could return `Vector` but that would lead to the same ambiguity discussed in [#84](https://github.com/AtheMathmo/rulinalg/pull/84).


src/matrix/base/mod.rs
```diff
@@ -211,6 +211,30 @@ pub trait BaseMatrix<T>: Sized {
         }
     }
 
+    /// Iterate over the columns of the matrix.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use rulinalg::matrix::{Matrix, BaseMatrix};
+    ///
```
AtheMathmo: Should use `matrix!` macro here

src/matrix/base/mod.rs
```diff
@@ -1321,6 +1350,32 @@ pub trait BaseMatrixMut<T>: BaseMatrix<T> {
...
     }
 
+    /// Iterate over the mutable columns of the matrix.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use rulinalg::matrix::{Matrix, BaseMatrixMut};
+    ///
+    /// let mut a = Matrix::new(3, 2, (0..6).collect::<Vec<usize>>());
```
AtheMathmo: And also `matrix!` here

src/matrix/iter.rs
```diff
@@ -580,6 +648,126 @@ mod tests {
         assert_eq!((0, Some(0)), diags_iter.size_hint());
     }
 
+    #[test]
+    fn test_matrix_cols() {
```
AtheMathmo: Need to add tests for `MatrixSlice(Mut)`.

c410-f3r:
Thanks @AtheMathmo and don't worry about time since I'm not in a hurry.

AtheMathmo:
Just had another check through. We still need to add tests for matrix slices (which I strongly suspect will fail without `row_stride`).

c410-f3r:
Did a little illustration to demonstrate my understanding about row_stride.

sans titre

`MatrixSlice::from_raw_parts(ptr, 3, 2, 2)`

Is it right?

AtheMathmo:
The illustration is neat! I wish I could put in as much effort to explain it.

The `row_stride` in the above matrix would be 4 - it is sort of the column count of the underlying data structure. The idea is that it tells us how far we need to go when we reach the end of the slice row, to get to the start of the next slice row.

To use your example, if we are at the start of the first slice row, so on element `6`, then we go four steps through the data: `7`, `8`, `9`, `10` to get to the start of the next slice row.

Does that make sense?

tafia:
Saying it differently, each row *slice* belongs to one matrix row. These matrix rows (owned) represent chunks of a continuous vector (row major storage), each having a length equals to row_stride. 

This info is needed if you want to navigate in your matrix slice as you are actually navigating in the matrix owned vector.

Not sure I am not adding more confusion, English is not my mother language.

Andlon:
I think this discussion suggests that we may want to provide a clear explanation of the storage format for matrices and matrix slices in the docs ;-)

AtheMathmo:
Maybe a wiki page would be most appropriate? If not then a separate `DEVELOPER_NOTES.md` file in the root? What would you suggest?

Andlon:
Well, considering that we expose details about the internal representation (i.e. through `data()`, `row_stride()` etc.), I think this should be described in the docs, presumably in the documentation for `BaseMatrix`? For end-users, knowing how the data is laid out is important not just for understanding, but also for interopability with other linear algebra libraries (including C libraries through FFI).

c410-f3r:
I guess this is it. Thanks everybody for helping me out.

src/matrix/base/mod.rs
```diff
@@ -1321,21 +1358,56 @@ pub trait BaseMatrixMut<T>: BaseMatrix<T> {
...
+    /// Iterate over the mutable columns of the matrix.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # #[macro_use] extern crate rulinalg; fn main() {
+    /// use rulinalg::matrix::{Matrix, BaseMatrixMut};
+    ///
+    /// let mut a = matrix![0, 1;
+    ///                 2, 3;
```
AtheMathmo: It looks to me like the indentation here is off. I'll have to check by rendering docs to be sure.

src/matrix/base/mod.rs
```diff
@@ -1321,21 +1358,56 @@ pub trait BaseMatrixMut<T>: BaseMatrix<T> {
...
     ///
     /// # Examples
     ///
     /// ```
+    /// # #[macro_use] extern crate rulinalg; fn main() {
     /// use rulinalg::matrix::{Matrix, BaseMatrixMut};
     ///
-    /// let mut a = Matrix::new(3, 2, (0..6).collect::<Vec<usize>>());
+    /// let mut a = matrix![0, 1;
+    ///                 2, 3;
```
AtheMathmo: And indentation here too.

src/matrix/iter.rs
```diff
@@ -580,6 +648,175 @@ mod tests {
...
+        let mut a = matrix![0, 1, 2, 3;
+                            4, 5, 6, 7;
+                            8, 9, 10, 11];
+        let data = [[0, 4, 8], [1, 5, 9], [2, 6, 10], [3, 7, 11]];
+
+        for (i, col) in a.col_iter().enumerate() {
+            for (ii, value) in col.iter().enumerate() {
```
AtheMathmo: **Very minor** but I find this `ii` index variable a little unconventional. If you have to make other changes could you swap to `j`, or `k`?

AtheMathmo: Actually - you should be able to use `assert_matrix_eq!` here to lighten the load a little. Not super important though.

src/matrix/iter.rs
```diff
@@ -580,6 +648,175 @@ mod tests {
...
+            }
+        }
+
+        assert_eq!(a.into_vec(), vec![0, 0, 2, 0, 0, 5, 6, 7, 8]);
+    }
+
+    #[test]
+    fn test_matrix_cols_nth() {
+        let a = matrix![0, 1, 2, 3;
+                            4, 5, 6, 7;
```
AtheMathmo: Indentation is off here.

AtheMathmo:
This looks ready to merge to me. Will give @tafia / @Andlon 48 hours in case they want to take a look. If not I'll get it in after.

src/matrix/iter.rs
```diff
@@ -107,6 +106,75 @@ impl<'a, T, M: $diag_base<T>> ExactSizeIterator for $diag<'a, T, M> {}
...
+            column = $col_base {
+                col: $slice_base::from_raw_parts(ptr, self.slice_rows, 1, self.row_stride as usize)
+            }
+        }
+        self.col_pos += n + 1;
+        Some(column)
+    }
+
+    fn count(self) -> usize {
+        self.slice_rows - self.col_pos
```
Andlon: Should this not be `self.slice_cols()`? I've only briefly skimmed the code, however, so maybe I misunderstand.

If so, this suggests you might want to test iterating over a non-square matrix to completion (which would catch this discrepancy) instead of a square matrix (in which case this bug is not apparent).

src/matrix/iter.rs
```diff
@@ -107,6 +106,75 @@ impl<'a, T, M: $diag_base<T>> ExactSizeIterator for $diag<'a, T, M> {}
...
+    }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        (self.slice_rows - self.col_pos, Some(self.slice_rows - self.col_pos))
```
Andlon: Same as in `count()`?

src/matrix/base/mod.rs
```diff
@@ -211,19 +211,51 @@ pub trait BaseMatrix<T>: Sized {
...
-    /// let a = Matrix::new(3, 2, (0..6).collect::<Vec<usize>>());
+    /// let a = matrix![0, 1;
+    ///                 2, 3;
+    ///                 4, 5];
+    ///
+    /// // Prints "3" two times.
+    /// for col in a.col_iter() {
+    ///     println!("{}", col.rows());
+    /// }
+    /// # }
```
AtheMathmo: I agree 100%. I was tempted to add a similar comment myself but as I am responsible for the previous poor examples I thought it was in bad taste :)

I'm happy for @c410-f3r to ignore this and I will improve the examples either before merging this or in a new PR.

AtheMathmo:
Thank you for the review @Andlon - you noticed some very important things!

As suggested feel free to leave the problems that affect both row and column iteration and I can handle these in a new PR after merging this.

bluss:
ndarray's illustrations are not very pretty, but they are at least home made so I can tell you to copy or use them without problem (any/no license or attribution needed).

https://bluss.github.io/rust-ndarray/talk1/rep-stride.svg
https://bluss.github.io/rust-ndarray/talk1/rep-ndarray.svg

AtheMathmo:
@bluss awesome, thank you!

c410-f3r:
Thanks everybody. I didn't modify the whole test scenario since this deserves its own PR.

AtheMathmo:
I was bold and tried to resolve conflicts in the web editor... I'll fix them properly now.

Andlon:
What's the status of this PR? Is it ready to merge, or is there still something remaining?

AtheMathmo:
Sorry I forgot about this, I think it is good to merge. If you agree we can get it in now.

Andlon:
I just briefly skimmed over the changes since last time we reviewed. I think it looks good!

Issue #131, opened 2016-12-29
Permutation matrices


Andlon:
A [permutation matrix](https://en.wikipedia.org/wiki/Permutation_matrix) is a very special kind of matrix with many applications in mathematics. Within rulinalg, its use is most notable in the LU decomposition: an `n x n` matrix `A` is decomposed such that `PA = LU` where `P` is an `n x n` permutation matrix.

Currently, the permutation matrix returned by our LU decomposition is an ordinary `Matrix` instance, which means it takes O(n<sup>2</sup>) storage and performing the multiplication `P B` for an `n x n` matrix `B` takes O(n<sup>3</sup>) operations.

However, because of its very special structure, a permutation matrix can be stored in O(n) storage, and the multiplication `P B` can be performed in only O(n<sup>2</sup>) operations. In my opinion, these cost savings more than justify having a specialized implementation for permutation matrices.

Most of the implementation of such a permutation matrix is already in place as part of #94. However, I intend to split up #94 so that the permutation matrix implementation can be a separate PR, which can be merged without breaking changes, whereas the full LU re-implementation may have to wait until we want to make a version bump with breaking changes.

I believe the following operations involving permutation matrices need to be supported:

- [x] Construction from an array-representation of a permutation
- [x] Computing the inverse permutation
- [x] Left-multiplying any `BaseMatrix` derivatives (permuting the rows)
- [x] Right-multiplying any `BaseMatrix` derivatives (permuting the cols)
- [x] Multiplication with other permutation matrices
- [x] Functionality to convert into a full `Matrix` (`From` implementation for `Matrix` presumably, or perhaps also a shortcut member function in addition)
- [x] Computing the parity/determinant of the permutation matrix
- [x] `rows()`/`cols()` (replaced with `size()` in #139)
- [x] Swapping rows in the permutation matrix
- [x] NO support for `Add`/`Sub` binary/unary operators (does not make sense for a permutation matrix)

I'm a little apprehensive about adding more matrix types for which multiplication implementations need to be provided, but after having thought about it a lot, I truly believe it is worth the effort for permutation matrices.

Also, the above suggests that perhaps we should have a more general matrix trait on which `BaseMatrix` depends. I.e. something like `MatrixLike` which covers only indexing and `rows`/`cols`.

AtheMathmo:
This is very well written, thank you!

I agree with pretty much everything you said. Just a couple comments:

[...]

PR #139, opened 2017-01-07
Permutation matrices

Andlon:
The following is an (almost complete) implementation of #131. Compared with the list in #131, what remains is:

- Products between permutation matrices
- Determinant/parity computation for permutation matrices

In addition, I have a few questions:

- So far I have not implemented multiplication for `MatrixSlice(Mut)`, only `&MatrixSlice(Mut)`. I see that i.e. `Mul<T>` is implemented for `MatrixSlice`. Consuming a matrix slice does not really make much sense in its own, but I suppose this is implemented so that the user does not have to deal with compilation errors when trying to multiply matrix slices and not their references? In that case I will also provide implementations also for `&MatrixSlice(Mut)`.
- For some of the `Px` mul implementations (`P` permutation matrix, `x` a `Vector`), I added some documentation detailing the complexity of each operation. I have not so far done it for the matrix-matrix products. Do you think this is worthwhile, or is the risk of the documentation getting out of sync too large for now, and it is something we could instead revisit once the library is more stable?

Also, note that I've separated all the logic of dealing with permutations into a very general `Permutation` data structure (which `PermutationMatrix` internally wraps). This way it's more focused and easier to test specifically only the permutation functionality. This leads to some duplication of APIs (and increased line count), but I think it makes the overall code cleaner. Currently this `Permutation` struct is exposed through the `utils` module, but in fact I want to propose making the whole `utils` module private (will make a separate issue for this eventually).

Finally, note that currently `PermutationMatrix` has a phantom marker that restricts it to a given type. In fact, one could implement and provide `PermutationMatrix` completely without any associations to a given scalar type (since it only deals with mapping indices). Put differently, it's possible to make `PermutationMatrix` _not_ a generic type, and rather a single implementation which could interact with `Matrix<T>` for any `T`. However, since `BaseMatrix` is generic over the scalar field rather than having the scalar field as an associated type, one cannot write functions that operate on `PermutationMatrix` and `BaseMatrix<T>` without also associating `PermutationMatrix` with `T`, hence the phantom marker. 

@AtheMathmo: This is one of the reasons I mentioned the possibility of rewriting `BaseMatrix` to use an associated type for the scalar field rather than generic type parameter in some other issue that I can not currently find.

In any case, this is not a big issue as usually you'd just work with a single type in the first place.


Andlon:
Also note that I have not been very concerned with performance in this initial PR. The implementation is in general rather simple, using only simple data structures and iteration. I'd expect it to be reasonably fast, but as no area of `rulinalg` currently uses `PermutationMatrix`, I do not have any reference points to work with. I'd say we rather get the APIs and functionality right, and then we can possibly improve performance in later PRs (possibly as part of a rewritten #94).

Andlon:
Just a heads-up: I think I will try to refactor the code so that I can avoid some of the duplication with `Permutation`. That is, I'll outright remove `Permutation` and leave `PermutationMatrix` behind, but I'll try to make sure that the permutation logic can still easily be tested on its own.

Just letting you know so that perhaps any sort of detailed review should wait a little. Also, the times I have to work on this are rather unpredictable, so progress may move at an erratic pace.

Andlon:
All right, I consider this ready for review now.

I added `itertools` as a dev-dependency. We've briefly discussed it before even for a general dependency. Adding it as a dev dependency means we can leverage it for tests, at least. If you don't want it, I can rewrite the test in which I use it, but I have to say that the test is a lot more understandable with it.

I realize it might be hard to review this code if you're not familiar with working with permutations. However, I've added ample tests and I've tried to use QuickCheck to cover the majority of the input space. So hopefully reviewing the tests should convince you that the implementations are most likely correct (unless I've forgotten to include relevant tests).

AtheMathmo:
Thank you for your work on this!

I'll take some time to look through it tomorrow. I've started getting a little more settled so should be able to put some more time into this soon.

AtheMathmo:
Before I review, I'll address some of the questions in the initial post. Sorry that I missed these (I thought we had discussed some before...)

> In that case I will also provide implementations also for `&MatrixSlice(Mut)`.

As you pointed out this is messy and for `MatrixSlice` not really necessary at all (it derives `Copy`). However I think that for now we should implement for references (as you have) just to save the compilation errors.

> [On documenting complexity] Do you think this is worthwhile, or is the risk of the documentation getting out of sync too large for now [...]?

I think that this could be worthwhile even now, but honestly may not be the best use of your time. It will be a little tricky to maintain and the existing documentation you have is plenty. It's up to you but I'm more than happy for you to stick with what you have.

> This is one of the reasons I mentioned the possibility of rewriting `BaseMatrix` to use an associated type 

I think it is a pretty compelling reason. Thanks for pointing it out.

Ok now I'll try to read the actual code :)

src/matrix/permutation_matrix.rs
```diff
@@ -0,0 +1,838 @@
...
+///
+/// ```text
+/// p: { 1, ..., n } -> { 1, ..., n }.
+/// ```
+///
+/// That is, for any index `i`, the permutation `p` sends `i` to some
+/// index `j = p(i)`, and hence the map may be represented as an array of integers
+/// of length *n*.
+///
+/// # Examples
```
AtheMathmo: I wonder if this example should appear a little earlier? I imagine most users will want to see an example usage before the theoretical and computational discussion (which is very well written, by the way).

src/matrix/permutation_matrix.rs
```diff
@@ -0,0 +1,838 @@
...
+/// let mut p = PermutationMatrix::identity(3);
+/// p.swap_rows(0, 1);
+/// assert_eq!(expected, p * x);
+///
+/// // Swap the two last columns of x by right-multiplying a permutation matrix
+/// let expected = matrix![1, 3, 2;
+///                        4, 6, 5;
+///                        7, 9, 8];
+/// let mut p = PermutationMatrix::identity(3);
+/// p.swap_rows(1, 2);
```
AtheMathmo: Is it worth explaining why we call `swap_rows` here despite permuting columns? I think it's safe to expect users to follow this.

src/matrix/permutation_matrix.rs
```diff
@@ -0,0 +1,838 @@
...
+    /// Swaps rows i and j in the permutation matrix.
+    pub fn swap_rows(&mut self, i: usize, j: usize) {
+        self.perm.swap(i, j);
+    }
+
+    /// The inverse of the permutation matrix.
+    pub fn inverse(&self) -> PermutationMatrix<T> {
+        let mut inv: Vec<usize> = vec![0; self.size()];
+
+        for (source, target) in self.perm.iter().cloned().enumerate() {
```
AtheMathmo: Do we need to call `cloned` here with `usize` implementing `Copy`? Also - we might get some very minor performance gains here [from unsafe code](https://github.com/AtheMathmo/rulinalg/blob/bb3e02297b33c4e7260dd226644d87e8abd11c5b/src/matrix/impl_mat.rs#L288-L300). This is premature but I thought I'd point it out (this is only safe with `Copy` types, I can't remember exactly why, something to do with `Drop`).

src/matrix/permutation_matrix.rs
```diff
@@ -0,0 +1,838 @@
...
+            perm: array.into(),
+            marker: std::marker::PhantomData
+        };
+        validate_permutation(&p.perm).map(|_| p)
+    }
+
+    /// Constructs a `PermutationMatrix` from an array, without checking the validity of
+    /// the supplied permutation.
+    ///
+    /// However, to ease development, a `debug_assert` with regards to the validity
```
AtheMathmo: Do you think it might be more conventional to remove the `debug_assert` and make this `unsafe`?

Right now I think that an invalid `perm` array wont lead to undefined behaviour, but it is possible this might be true in the future (I'm thinking of the `inverse` function in particular).

src/matrix/permutation_matrix.rs
```diff
@@ -0,0 +1,838 @@
...
+    pub fn from_array_unchecked<A: Into<Vec<usize>>>(array: A) -> PermutationMatrix<T> {
+        let p = PermutationMatrix {
+            perm: array.into(),
+            marker: std::marker::PhantomData
+        };
+        debug_assert!(validate_permutation(&p.perm).is_ok(), "Permutation is not valid");
+        p
+    }
+
+    /// Maps the given row index into the resulting row index in the permuted matrix.
```
AtheMathmo: I don't think this function is obvious from the description. Maybe add a small example?

```rust
let a = PermutationMatrix::identity();
a.swap_rows(2,3);
assert_eq!(a.map_row(2), 3);
```

src/matrix/permutation_matrix.rs
```diff
@@ -0,0 +1,838 @@
...
+/// container.
+///
+/// The permutation is applied by calls to `swap(i, j)` for indices
+/// `i` and `j`.
+///
+/// # Complexity
+///
+/// - O(1) memory usage.
+/// - O(n) worst case number of calls to `swap`.
+fn permute_by_swap<S>(perm: &mut [usize], mut swap: S) where S: FnMut(usize, usize) -> () {
```
AtheMathmo: Just wanted to say this function is really neat! It makes a lot of the above code clean.

src/matrix/permutation_matrix.rs
```diff
@@ -0,0 +1,838 @@
...
+
+    #[test]
+    fn permute_vector_in_place() {
+        let mut x = vector![ 0, 1, 2, 3];
+        let p = PermutationMatrix::from_array(vec![2, 1, 3, 0]).unwrap();
+        p.permute_vector_in_place(&mut x);
+        assert_vector_eq!(x, vector![ 3, 1, 0, 2]);
+    }
+
+    use quickcheck::{Arbitrary, Gen};
```
AtheMathmo: I wonder if this and below should be placed in the `tests` folder and treated as regression tests. I don't mind much either way. (I'm glad we added `quickcheck` by the way, these tests are really cool).

src/matrix/permutation_matrix_mul.rs
```diff
@@ -0,0 +1,416 @@
...
+
+/// Left-multiply a vector by a permutation matrix.
+impl<'a, T> Mul<Vector<T>> for &'a PermutationMatrix<T> where T: Clone + Zero {
+    type Output = Vector<T>;
+
+    fn mul(self, rhs: Vector<T>) -> Vector<T> {
+        // Here we have the choice of using `permute_by_copy`
+        // `permute_by_swap`, as we can reuse one of the existing
+        // implementations.
+        self * &rhs
```
AtheMathmo: Sorry if I am misreading. It looks like here we defer to the case where we have a reference to the `Vector<T>`. But in this case we can permute the vector in place (as in the implementation above).

src/matrix/permutation_matrix.rs
Andlon: Yes, I completely agree! I will reorder the docs here.

src/matrix/permutation_matrix.rs
Andlon: I can see how this may seem a little confusing. Perhaps the explanation could emphasis more on the fact that we represent a representation matrix in terms of permutations of the row indices of a matrix? (This is also conventional)

I'm not entirely sure how to make it more clear though, but perhaps a short comment in the example would help.

src/matrix/permutation_matrix.rs
Andlon: IIRC, the reason is because indexing is defined in terms of `usize`, not `&usize`, so one does actually have to make a copy (but for primitive types this is of no consequence performance-wise anyway, the compiler will make the best of it).

Yes, performance can likely be increased by avoiding the allocation as you point out. I was contemplating it as I wrote it, but I came to the conclusion that I'd rather write a perfectly safe function now, and then possibly improve performance in a later PR if is necessary (i.e. LU).

src/matrix/permutation_matrix.rs
Andlon: Ah, good catch! I didn't think of it this way. Yes, indeed, if we decide to use unsafe indexing in _any_ of the other methods, then UB might be invoked. I agree, this function should be marked `unsafe`, and perhaps also emphasis the fact that this function can cause undefined behavior in *other* methods on `PermutationMatrix`

src/matrix/permutation_matrix.rs
Andlon: Totally agree!

src/matrix/permutation_matrix.rs
Andlon: Yeah, I'm a little conflicted here. I'm OK with either way, I'll leave the final decision up to you. My way of thinking is this:

- In general, it's nice to have tests for isolated behavior close to the implementation of said behavior. So, something that only tests `PermutationMatrix` would be nice to have close to the `PermutationMatrix` implementation. So, unit- and property tests for a single piece of functionality, such as `PermutationMatrix` makes sense to put close to the implementation.
- If you test more than a single unit, they are by definition integration tests, and that's what the `tests` folder is intended for (according to https://doc.rust-lang.org/book/testing.html). For mathematical libraries, it can sometimes be a bit hard to distinguish unit and integration tests. For example, if you want to check that the orthogonal matrices produced by SVD have determinant 1 or -1, would this be a unit test for SVD (which implicitly depends on LU for determinant), or would it be an integration test? You *could* make the case for both, although I think I lean towards the latter.

Just so we're on the same page, the term "regression test" for me can be any kind of test - unit, integration, whatever. The characteristic property of a regression test is that it *verifies that something that previously broke is now working*. For example, if a user creates an issue and claims that our LU decomposition fails for the `2 x 2` identity matrix, we would first create a unit test from this and verify that it fails. Next, we would fix the LU decomposition and check that the unit test passes. Then we keep the test as a regression test (to make sure that this *particular* bug never comes back). With this in mind, I'd say that a "regression test" can be either a simple unit test (which you might place in the same file as the implementation) or an integration test (which you might put in the `tests/` folder).

This became a bit longer than I had planned. I hope I don't come off as lecturing. I just wanted to share parts of my philosophy with regards to testing!

src/matrix/permutation_matrix_mul.rs
Andlon: Yeah, I get your confusion. The comment was already confusing when I wrote it, but after I made some changes it's completely out of sync and doesn't make sense anymore.

The problem is that, while we take the `Vector` by value, we only have a reference to the permutation matrix. In order to do an in-place permutation, we actually need to take the permutation matrix by value. So in both cases, we'd need to allocate O(n) memory, and it doesn't really matter too much what choice we make.

I suspect it might be a better choice to clone the `PermutationMatrix` and permute the vector in-place instead, because currently, when the vector is permuted out-of-place, we first zero the newly allocated buffer (due to not using unsafe code as in the case of `inverse()`).

I'll change it and update the comment to try to better explain what's going on :)

Andlon:
> As you pointed out this is messy and for MatrixSlice not really necessary at all (it derives Copy). However I think that for now we should implement for references (as you have) just to save the compilation errors.

Heh, I made a typo there. At the moment I'm actually missing the implementations for the value types. I'll add them when I get the chance.

> I think that this could be worthwhile even now, but honestly may not be the best use of your time. It will be a little tricky to maintain and the existing documentation you have is plenty. It's up to you but I'm more than happy for you to stick with what you have.

I think that while we should start documenting complexity in general, I think it's not necessarily the best usage of time, because rulinalg as a whole is still very volatile. However, in the future, once things are more stable and the algorithms are more battle-tested, we should perhaps start documenting these things. For now I decided to leave it out.

Andlon:
> Would you be able to annotate areas of code that you know can be improved (performance-wise)? I just mean if there is anything you came across in your research but that was too hefty for a first implementation it would be good to know.

Well, apart from more efficient indexing, there are no low-hanging fruits that I'm aware of. Unlike e.g. matrix-matrix products, the performance of permutation matrices depend highly on the data itself. For example, applying the identity matrix in place is basically equivalent to iterating over `n` items, while a worst-case shuffle might involve moving every element in the cache-wise worse possible way.

In fact, I believe that it is *proven* that there exists no cache-optimal algorithm for applying a permutation. One can however improve cache performance in most cases by representing a permutation by two arrays instead of one, and then sorting them in some specific way, although I never saw an explanation of exactly how they should be sorted. Still, this seems more like a theoretical thing, because the costs of sorting them and having two arrays add *significant* overhead in practice, which means that it's not universally faster, which means you'd have to apply it conditionally depending on some heuristics. Ugh, it wouldn't be pretty, at least.

To summarize, I'm not sure exactly where to put this in the code, but this is at least my impression of performance characteristics. Bear in mind, however, that I'm certainly no expert on this, I've only skimmed a few papers.

> Very minor: I wonder if to stick with the theme permutation_matrix_mul.rs should be impl_permutation_mul.rs?

Yes, I can certainly make that change :)

> I think a condition for the merge of this PR should be to make an issue taking note of the areas in code where we would like to introduce this new type (lup being a big one).

Agree. Though, to be fair, I can't think of any other applications than LU right now, off the top of my head. They come up quite often when working with sparse direct solvers, however, so we might see more use of them in the future.

Actually, there's one more application in the current code base. I've known for some time that I actually incorrectly implemented the sorting of the SVD values. I'll have to get around to making an issue for that (or add it to the existing SVD issue). Turns out, this should be possible to trivially fix with the new `PermutationMatrix` type.

Thanks a lot for your insightful comments and suggestions! I'll try to get around to fixing these things during the week :)

AtheMathmo:
> Heh, I made a typo there. At the moment I'm actually missing the implementations for the value types. I'll add them when I get the chance.

No worries! This is certainly minor, I thought it was useful mostly for consistency.

Thank you for explaining the performance a little more. The reason I asked is that I couldn't identify anything myself and didn't want to miss anything in the future. Things look great to me so don't worry about any minor theoretical performance gains :)

In addition to LUP and SVD there is a `parity` function in `matrix/mod.rs` that we can remove once this is merged. I'm happy to write up this issue myself once we merge.

src/matrix/permutation_matrix.rs
AtheMathmo: I think that stating the convention that we're adhering to might be worthwhile. That and/or a short comment.

This is very minor anyway, so don't worry about it too much.

src/matrix/permutation_matrix.rs
AtheMathmo: Sounds good. We can clarify why the function is `unsafe` in the docs `Safety` section.

src/matrix/permutation_matrix_mul.rs
AtheMathmo: Ah right, I understand now. Thanks!

The other thing to consider is the Vector data type. If our `usize` is a 64 bit pointer and our data type is 32 bit (i.e. `f32`) then we would probably prefer to clone the `Vector`. Otherwise larger custom data types, say `Vector<[u64; n]>` would lead us to clone the permutation matrix.

I'd say this is out of scope for now (and perhaps irrelevant?) and we should go with your gut and clone the permutation.

src/matrix/permutation_matrix.rs
AtheMathmo: > Just so we're on the same page, the term "regression test" for me can be any kind of test - unit, integration, whatever.

Ah yes, sorry I did mean to write integration.

Thanks for the thorough explanation. I think that now I do actually agree with you. It makes more sense to keep these tests here.

Part of my motivation for the opposite argument was that I really liked how you built the generators to create valid and invalid permutation matrices. I thought that in the future it might be useful to create a suite of similar tools in our `tests` folder to cover a wider range of algebraic properties. For example generating singular matrices (or matrices of rank k), (./bi/tri)diagonal matrices, etc.

Andlon: This is, in fact, *exactly* what I have in mind :) I am planning to write a blog post at some point on how one can use property-based testing for linear algebra, and especially how we use it in rulinalg (well, how we will use it). Like you say, it's quite easy to generate i.e. singular matrix, matrices of a certain rank, arbitrary well-conditioned matrices or even ill-conditioned matrices just by tweaking e.g. eigenvalues or singular values and basically doing a reverse decomposition (such as constructing the eigenvalues first, and then use random orthogonal matrices to construct an arbitrary matrix with said eigenvalues).

I think in general, for some type that exists in the library, it makes sense to put the implementation of `Arbitrary` in the same file, in the `tests` module. I think then *the implementation* should still be accessible from the `tests` folder, shouldn't it?

In addition, I want to create something like a `test_support` module (which is only available in `cfg(test)`), which includes structs such as `SingularMatrix`, `WellConditionedMatrix`, `InvertibleMatrix`, `OrthogonalMatrix` etc. and their corresponding `Arbitrary` implementations.

AtheMathmo: > I am planning to write a blog post at some point on how one can use property-based testing for linear algebra, and especially how we use it in rulinalg (well, how we will use it).

I think this is an excellent idea! I believe it's a cool idea that will resonate with members of the wider community.

We have time to decide on the exact structure, but I think what you suggest here makes sense. We will be able to import from the `tests` module, i.e. `decomposition::svd::tests::SingularMatrix`. A `test_support` module would also be a good idea. Another idea might be to just have a separate micro crate within the repository for these tools? We can discuss this in more detail later though.

Andlon:
All right, unless I missed anything, I believe I've addressed the points you brought up!

src/matrix/permutation_matrix.rs
```diff
@@ -0,0 +1,885 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use matrix::Matrix;
+    use vector::Vector;
+    use super::{PermutationMatrix, Parity};
+    use super::{permute_by_swap, validate_permutation};
+
+    use itertools::Itertools;
```
AtheMathmo: One question though - would you mind explaining how this works and what it affects?

PR #141, opened 2017-01-29
Fix Matrix min / max impl

sinhrks:
Let me correct my misunderstanding... output shouldn't be changed. (#115)


AtheMathmo:
Well spotted, thanks!

Andlon:
Just want to point out that I think the `min`/`max` functions may panic for empty matrices (not due to this PR, I just glanced over it and saw it in the existing code)?

(also, can a matrix have dimensions `0 x n` or `n x 0` for `n` non-zero? If so this may also cause problems)

PR #146, opened 2017-02-03
Fix SliceIter overflow bug in corner cases

Andlon:
It turns out that `SliceIter` does not play nice for matrices that have dimension `(k, 0)` for `k > 0`. In particular, it overflowed a `usize` because in this case it computed `0usize - 1usize`. I had to rewrite the logic a little bit in order to avoid an additional bounds check for the column (hopefully the compiler will optimize `end` as a constant instead of recomputing it every call to `next`). I also couldn't really find any tests for `iter()`, so I added some, although I'm not sure if they perhaps actually exist somewhere.

### DIFF
--- a/src/matrix/iter.rs
+++ b/src/matrix/iter.rs
@@ -767,4 +767,22 @@ mod tests {
             }
         }
     }
+
+    #[test]
+    fn iter_empty_matrix() {
+        {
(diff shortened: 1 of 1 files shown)


AtheMathmo:
This is a good spot! Thank you.

I thought some tests did exist somewhere but I cannot find them. It's a good job you spotted that too!

Issue #151, opened 2017-02-14
BaseMatrix has no `get` method.


Andlon:
As the title says, `BaseMatrix` has no `get` method, only `get_unchecked`. Is this deliberate or just an omission? I'm rewriting some generic code (mainly forward and back substitution) and I wanted to avoid unsafe code.

AtheMathmo:
This is an ommission, thanks for pointing it out! We might also want to remedy this by requiring an `Index<[usize; 2>` trait bound on `BaseMatrix`? I think the `get` matrix function is a good addition regardless.

prismatica:
Is anyone working on this? If you're looking for help, this looks like a good place to get started!

Andlon:
@prismatica: not to my knowledge. You're welcome to give it a go if you want :)

AtheMathmo:
@prismatica please take it!

If you need any help locating things or implementing the code feel free to comment here.

PR #152, opened 2017-02-14
Rewrite forward_substitution and back_substitution

Andlon:
I had some problems with these functions because they returned an error for empty matrices. In my opinion, solving a system `Ax = b` for empty `A` and `b` should merely return an empty `x`. I've had some major grievances in the past with this exact situation with Eigen, which doesn't handle this situation very well. My application generated linear systems from geometrical meshes, and sometimes these systems turned out to be "empty". Graceful handling is thus important so that users don't need to special-case these scenarios.

In any case, I took the liberty to completely rewrite the forward and backward substitution functions. I used some of the techniques I discovered when working on #150 to rewrite them in terms of the highly optimized `utils::dot` function, as well as leveraging our row iterators. They no longer contain any unsafe code, except for needing to call `get_unchecked(i, i)` from `BaseMatrix`. See #151.

The resulting code is significantly simpler, does not allocate a new vector like the previous implementation and is also significantly faster. Here are the benchmarks **before** making the changes:

```text
test linalg::triangular::solve_l_triangular_10000x10000 ... bench:  51,194,669 ns/iter (+/- 323,709)
test linalg::triangular::solve_l_triangular_1000x1000   ... bench:     559,368 ns/iter (+/- 8,496)
test linalg::triangular::solve_l_triangular_100x100     ... bench:       5,394 ns/iter (+/- 550)
test linalg::triangular::solve_u_triangular_10000x10000 ... bench:  50,927,341 ns/iter (+/- 879,902)
test linalg::triangular::solve_u_triangular_1000x1000   ... bench:     550,613 ns/iter (+/- 12,704)
test linalg::triangular::solve_u_triangular_100x100     ... bench:       4,708 ns/iter (+/- 17)
```

Here are the benchmarks **after** making the changes:

```text
test linalg::triangular::solve_l_triangular_10000x10000 ... bench:  29,930,075 ns/iter (+/- 263,948)
test linalg::triangular::solve_l_triangular_1000x1000   ... bench:     296,419 ns/iter (+/- 10,734)
test linalg::triangular::solve_l_triangular_100x100     ... bench:       2,574 ns/iter (+/- 3)
test linalg::triangular::solve_u_triangular_10000x10000 ... bench:  30,100,720 ns/iter (+/- 325,360)
test linalg::triangular::solve_u_triangular_1000x1000   ... bench:     295,012 ns/iter (+/- 3,668)
test linalg::triangular::solve_u_triangular_100x100     ... bench:       3,188 ns/iter (+/- 4)
```

### DIFF
--- a/tests/mat/mod.rs
+++ b/tests/mat/mod.rs
@@ -1,4 +1,4 @@
-use rulinalg::matrix::{BaseMatrix, Matrix};
+use rulinalg::matrix::{BaseMatrix};
 
 #[test]
 fn test_solve() {
@@ -33,9 +33,6 @@ fn test_l_triangular_solve_errs() {
 
 #[test]
 fn test_u_triangular_solve_errs() {
-    let a: Matrix<f64> = matrix![];
-    assert!(a.solve_u_triangular(vector![]).is_err());;
-
     let a = matrix![0.0];
     assert!(a.solve_u_triangular(vector![1.0]).is_err());
 }
--- a/tests/mat/mod.rs
+++ b/tests/mat/mod.rs
@@ -15,17 +15,18 @@ fn test_solve() {
     let b = vector![-100.0, 0.0, 0.0, -100.0, 0.0, 0.0, -100.0, 0.0, 0.0];
 
     let c = a.solve(b).unwrap();
-    let true_solution = vec![42.85714286, 18.75, 7.14285714, 52.67857143,
-                             25.0, 9.82142857, 42.85714286, 18.75, 7.14285714];
+    let true_solution = vector![42.85714286, 18.75, 7.14285714, 52.67857143,
+                                25.0, 9.82142857, 42.85714286, 18.75, 7.14285714];
 
-    assert!(c.into_iter().zip(true_solution.into_iter()).all(|(x, y)| (x-y) < 1e-5));
-
+    // Note: the "true_solution" given here has way too few
+    // significant digits, and since I can't be bothered to enter
+    // it all into e.g. NumPy, I'm leaving a lower absolute
+    // tolerance in place.
+    assert_vector_eq!(c, true_solution, comp = abs, tol = 1e-8);
 }
 
 #[test]
 fn test_l_triangular_solve_errs() {
-    let a: Matrix<f64> = matrix![];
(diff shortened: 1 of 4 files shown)


Andlon:
Update: I took the liberty of removing the `Any` trait bounds from these functions too. AFAIK, removing a trait bound is not a breaking change...?

AtheMathmo:
This is awesome! I don't think that removing `Any` is a breaking change either. I'll try to check this over properly later today. (It's a busy week for me with a midterm)

Andlon:
@AtheMathmo: please take as long as you want with the reviews. I'd rather you focus properly on your midterm :) I just found myself with some unexpected time to work on `rulinalg`, because my current temporary situation precludes me from doing my actual thesis work.

AtheMathmo:
I massively appreciate all of the changes you've put in so far - they're big leaps in usability and performance. Thank you!

Andlon:
Thanks for saying so! I'm learning a lot myself in the process, so I'm very happy to work on it :)

AtheMathmo:
I merged in #142 which has given some conflicts here. I'll try to review regardless.

tests/mat/mod.rs
```diff
@@ -15,26 +15,24 @@ fn test_solve() {
...
     let c = a.solve(b).unwrap();
-    let true_solution = vec![42.85714286, 18.75, 7.14285714, 52.67857143,
-                             25.0, 9.82142857, 42.85714286, 18.75, 7.14285714];
-
-    assert!(c.into_iter().zip(true_solution.into_iter()).all(|(x, y)| (x-y) < 1e-5));
-
+    let true_solution = vector![42.85714286, 18.75, 7.14285714, 52.67857143,
+                                25.0, 9.82142857, 42.85714286, 18.75, 7.14285714];
+
+    // Note: the "true_solution" given here has way too few
```
AtheMathmo: This is definitely fine - I think if we are within 1e-8 we can safely assume the implementation is correct. We might be missing some tricks to improve accuracy but these should be captured in a separate test where necessary.

AtheMathmo:
This looks good to me. Once the conflicts are resolved I am happy to merge.

Andlon:
Should be good to go now assuming tests complete successfully. Thanks for taking a look!

Issue #132, opened 2016-12-29
Add a new `is_data_contiguous` function


AtheMathmo:
We have discussed around this before - it was most recently suggested in #86 .

I think this is a useful utility function that we could have on the `BaseMatrix` trait. The default implementation would check that the row stride and column count are the same (I think?). We would override the function for `Matrix` to always return true.

We could then use this check to improve our iterators and perhaps also a few other areas of code.

Issue #86, opened 2016-11-02
Iterator performance tips


bluss:
I'm briefly looking at rulinalg’s SliceIter. You can use the same kind of optimization used in ndarray's element iterator.

The brief idea of the "problem" is that this is efficent:

```
for i in rows
   for j in row(i)
       // access element at i, j
```

`SliceIter::next()` (just like ndarray's Iter::next) emulates this segmented traversal with a bit of a state machine. Unfortunately the compiler doesn't know how to turn this state machine back into two nested for loops(!)

Two main ideas I have used:

1. Use the regular `[T]` iterator when the matrix is contiguous. A branch on this even [inside the `.next()` method](https://github.com/bluss/rust-ndarray/blob/75aa62a57391eceb675248aa0d1bf7a9794a613b/src/iterators.rs#L281-L286) seems to work well, the conditional is lifted out of the loop.
2. When you can't improve .next() for the non-contiguous case, you can at least special case `Iterator::fold()`. It is essentially the same idea as the .chain() change in https://github.com/rust-lang/rust/pull/37315 ; you recurse and fold a bunch of component iterators (since each row is contiguous, you just call the slice iterator's fold on each row in turn).

AtheMathmo:
Thank you so much for opening this!

I think it will take me a little while to digest this, but from what I do understand it will be really valuable.

AtheMathmo:
I was just reading through this again @bluss . How do you differentiate between the contiguous and non-contiguous cases in ndarray? A contributor suggested we add specialist types for contiguous and non-contiguous data. This would help us pull out a little more performance in exchange for more compexity. I'm starting to come around to this idea and was wondering if you do something similar?

Thanks!

bluss:
ndarray computes this from the dimensions and strides every time the array's data is requested as slice or iterator. It's in the plan to cache this computation with a flag, but it hasn't been done yet.

I don't know how many different memory layouts your library supports, but it's fewer than ndarray, so it's possible that the check is a very cheap comparison.

bluss:
Type-level information lets you save code bloat, I'm sure that's useful. I didn't want to have more array types.

PR #153, opened 2017-02-15
Add a new `is_data_contiguous` function #132

epi16:
This resolves issue #132. It involved adding a new function to the
BaseMatrix trait called `is_data_contiguous` that returns true if the
elements of a matrix are stored contiguously.

### DIFF
--- a/benches/linalg/iter.rs
+++ b/benches/linalg/iter.rs
@@ -57,3 +57,31 @@ fn mat_diag_manual_100_500(b: &mut Bencher) {
         }
     });
 }
+
+#[bench]
+fn mat_contigous_iter_10_50(b: &mut Bencher) {
+
+    let a = Matrix::new(10, 50, vec![2.0;500]);
+    let slice = a.sub_slice([0,0], 10, 50);
+
+    b.iter(|| {
+        for x in slice.iter()
+        {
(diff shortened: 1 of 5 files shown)


AtheMathmo:
Awesome!

Thank you for getting this done. There are a few other PRs on my queue to review and merge first. This is fairly small so it should go in without any issues.

AtheMathmo:
I had a quick glance through the latest commit and it looks like a solid approach. I just wanted to check that you had a chance to read bluss' comments on the matter, in #86 . I haven't had a chance to confirm how he achieves this but if you haven't already it would definitely be worth looking at [ndarray](https://github.com/bluss/rust-ndarray/blob/75aa62a57391eceb675248aa0d1bf7a9794a613b/src/iterators.rs#L281-L286).

Thanks again by the way, it is good to get both of these done at once I think! :D

epi16:
Aaah! Sorry, I had tried to keep my most recent commit separate from your main repository. I'm just picking up GitHub, and a lot of the features are new to me. Irrespective, I've only skimmed ndarray's approach, and I doubt that what I actually did is production-ready.

AtheMathmo:
@epi16 did you mean to close this?

PR #157, opened 2017-02-22
Fix for documentation on `Matrix::from_fn()`

dyule:
This one caused me no small amount of grief until I look at the source (although the example is correct).  Ideally, it would be great to change the order of the function arguments from 'col', 'row' to 'row', 'col', but I imagine that would invisibly break a bunch of existing code (I know it would break mine).  You could potentially create a function function like `from_fn2()`, and deprecate the old one, but that's not super great either.

### DIFF
--- a/src/matrix/impl_mat.rs
+++ b/src/matrix/impl_mat.rs
@@ -41,7 +41,7 @@ impl<T> Matrix<T> {
     }
 
     /// Constructor for Matrix struct that takes a function `f`
-    /// and constructs a new matrix such that `A_ij = f(i, j)`,
+    /// and constructs a new matrix such that `A_ij = f(j, i)`,
     /// where `i` is the row index and `j` the column index.
     ///
     /// Requires both the row and column dimensions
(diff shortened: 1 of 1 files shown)


AtheMathmo:
Good spot!

I'm not sure whether we would prefer to update the docs or treat this as a bug and change the function itself. Of course this would be a breaking change.

I think I would prefer to change the function. Whether we do this in a minor or major version bump is up for debate. Any thoughts @Andlon ?

Andlon:
Huh. Yeah, this is completely unintuitive as it is right now. We definitely need to make it `(row, col)`. However, as @dyule rightly points out, this is a breaking change, and a rather nasty breaking change at that since it gives no compiler error - it merely breaks users' code in a semi-subtle way (depending on how good their test suite is).

Introducing a new function and deprecating this one is a good suggestion - however, `from_fn` is really the name we want to have.

Although I feel bad for the users whose code might be subtly broken by making this change, I think changing the function is the right thing to do. Rulinalg's API is still evolving at a rapid pace, and I think it's too early to let backward compatibility concerns have an impact on this evolution.

I'd say the best course of action might be to update the docs as @dyule has, but in addition state that *this will be changed* in a future release. Then we make the actual change in the next "major" version bump (0.5 I guess). I think this is slightly kinder than outright breaking their code in a minor version?

AtheMathmo:
Sorry for the delay!

I agree with @Andlon. I'll modify this PR to add a note that this function will be changed in a future release.
Then we'll merge as-is.

PR #158, opened 2017-02-25
Added "get" and "get_mut" to BaseMatrix and BaseMatrixMut

### LINKED ISSUE #151: BaseMatrix has no `get` method.

prismatica:
This resolves #151, unless you all also want to require an Index bound for BaseMatrix (which I didn't add).

### DIFF
--- a/src/matrix/base/mod.rs
+++ b/src/matrix/base/mod.rs
@@ -80,6 +80,35 @@ pub trait BaseMatrix<T>: Sized {
         &*(self.as_ptr().offset((index[0] * self.row_stride() + index[1]) as isize))
     }
 
+    /// Get a reference to a point in the matrix. 
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # #[macro_use] extern crate rulinalg; fn main() {
+    /// use rulinalg::matrix::{Matrix, BaseMatrix};
+    /// 
+    /// let mat = matrix![0, 1;
+    ///                   3, 4;
+    ///                   6, 7];
+    ///
+    /// assert_eq!(mat.get([0, 2]), None);
+    /// assert_eq!(mat.get([3, 0]), None);
+    ///
+    /// assert_eq!( *mat.get([0, 0]).unwrap(), 0)
+    /// # }
+    /// ```
+    fn get(&self, index: [usize; 2]) -> Option<&T> {
+				let row_ind = index[0];
+				let col_ind = index[1];
+				
+        if row_ind >= self.rows() || col_ind >= self.cols() {
+					None
+        } else {
+        	unsafe { Some(self.get_unchecked(index)) }
+        }
+    }
+
     /// Returns the column of a matrix at the given index.
     /// `None` if the index is out of bounds.
     ///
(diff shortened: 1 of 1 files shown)


src/matrix/base/mod.rs
```diff
@@ -80,6 +80,35 @@ pub trait BaseMatrix<T>: Sized {
...
+    ///
+    /// assert_eq!( *mat.get([0, 0]).unwrap(), 0)
+    /// # }
+    /// ```
+    fn get(&self, index: [usize; 2]) -> Option<&T> {
+				let row_ind = index[0];
+				let col_ind = index[1];
+				
+        if row_ind >= self.rows() || col_ind >= self.cols() {
+					None
```
AtheMathmo: Formatting here looks a little off - indentation is too large?

tafia:
No big deal and not directly related to this PR, but I was wondering if anyone tried adding some `#[inline]` attributes in these kind of functions?

src/matrix/base/mod.rs
```diff
@@ -80,6 +80,35 @@ pub trait BaseMatrix<T>: Sized {
         &*(self.as_ptr().offset((index[0] * self.row_stride() + index[1]) as isize))
     }
 
+    /// Get a reference to a point in the matrix.
```
Andlon: I feel that "point" is a little strange. However, I see that this is what the existing docs for `get_unchecked` say too, so you've certainly done the right thing by using the same language.

@AtheMathmo: Perhaps this should be changed to "an element" for both `get` and `get_unchecked` (and their `mut` brethren)?

Andlon:
@tafia: My understanding is somewhat limited in this regard, but I would expect it not to have any particular effect on generic functions, since they must be effectively "imported" into the compilation unit of the user anyway.

That is, my understanding of `#[inline]` is primarily that it tells the compiler that external crates may want to inline this function. However, for generic functions, I would expect the whole function to get monomorphized at the call site instead, in which case the compiler is able to inline it if it finds it appropriate.

Do take it with a large pinch of salt, however. I might be completely wrong here :)

src/matrix/base/mod.rs
prismatica: Yup, definitely! I can fix that tonight.

src/matrix/base/mod.rs
prismatica: I also think that "point" is unusual, for what it's worth. I'd be happy to update the comments, if that's the consensus.

AtheMathmo: Yeah let's change all of these to `element`. I'm not sure what I was thinking by `point` :)

AtheMathmo:
I'm also not sure if there is any benefit to inlining these functions. I've never compared inlining with generic functions but what @Andlon says makes sense.

Perhaps this is worth investigating separately to this PR though.

prismatica:
Are there any changes I should make to this latest merge?

AtheMathmo:
At a glance it looks good to me. I'll try to give it a proper look later tonight (I'm just about to head out).

Sorry for the delay - it's been a very busy week for me.

prismatica:
No problem at all, didn't mean to rush you!

PR #162, opened 2017-03-25
Sparse matrix

c410-f3r:
Lost track of the last PR, again. This new PR brings some of the features discussed in #83.

Guys, I am leaving this code as a record because I don't think it should be merged. My time now is dedicated to other projects and I won't be able to implement all the missing features (arithmetic, creations, iterators, methods and solvers), it is just too much and the complexity is way greater than dense matrices, e.g,. [http://crd-legacy.lbl.gov/~xiaoye/SuperLU/SparseDirectSurvey.pdf](http://crd-legacy.lbl.gov/~xiaoye/SuperLU/SparseDirectSurvey.pdf).

I will still continue contributing to the project with things more achievable and maybe return to this feature in the future.

Sorry for the inconvenience and thanks for all the help!

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -99,6 +99,7 @@ pub mod utils;
 pub mod vector;
 pub mod ulp;
 pub mod norm;
+pub mod sparse_matrix;
 
 mod internal_utils;
 
--- a/src/sparse_matrix/compressed_matrix/compressed/compressed_iter.rs
+++ b/src/sparse_matrix/compressed_matrix/compressed/compressed_iter.rs
@@ -1,0 +1,229 @@
+use std::slice;
+
+use sparse_matrix::{CompressedIter, CompressedIterMut};
+
+macro_rules! impl_iter (
+    ($iter_struct:ident, $data_type:ty, $slice_from_parts:ident) => (
+impl<'a, T> Iterator for $iter_struct<'a, T> {
+    type Item = (&'a [usize], $data_type);
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.current_pos >= self.positions {
+            return None;
+        }
+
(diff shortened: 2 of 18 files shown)


Andlon:
@c410-f3r: thank you for leaving the PR here. I think it's fair to say that we've all learnt a lot from your work on sparse matrices, and I think that it makes for an excellent starting point for further work.

The implementation of support for sparse matrices is a huge undertaking, and something I personally would absolutely love to eventually see implemented in `rulinalg`. I would also be interested in continuing your work myself at a later stage - however for the moment I'm mostly concerned with improving `rulinalg` for dense linear algebra.

I must reiterate that I've personally learnt a lot from your hard work, and I've gotten a much better idea of how we can handle the implementation of sparse matrices in the future. I'm also taking this into account when working on the dense portions, so that the integration of sparse data structures and algorithms will eventually be a smoother process (I hope).

You've already shown yourself as a very valuable contributor to the library, and I'd be very happy to see you contribute more!

AtheMathmo:
@c410-f3r thank you for taking the time to get this PR up to date.

It is indeed a huge undertaking and there's no need to apologise. I am in the same boat - I would have loved to finish up the previous PR but did not have the time to learn and implement the necessary code.

I hope that I will find some time to address this PR soon. I think that we were close to having enough functionality that it were worth adding this to the library - if this is the case then I expect we can make good use of your hard work still.

And as @Andlon said - we have all learnt a lot from this work and I see that as a very valuable contribution. Please don't be put off from handling other issues in the library - we'd welcome your help wherever you can give it!

c410-f3r:
I too learned a lot with you guys. Thanks for the good reception, now I am more relieved :)
Let's all of us move forward with more challenges ahead!

Issue #95, opened 2016-11-25
assert_matrix_eq! improvements (tracking issue)


Andlon:
In this issue, I intend to maintain a list of suggestions for improvements to the `assert_matrix_eq!` and `assert_vector_eq!` macros. As more contributors get a chance to use it, I would appreciate input on missing features that would make working with it easier. Actual bugs in these macros should still be reported in their own issue.

Potential usability improvements:
- [ ] We may want to consider implementing an `assert_scalar_eq!` for scalar comparisons, leveraging the same comparators. Currently we have no facility for this in `rulinalg`, and it's a situation that comes up a lot in tests (such as testing that the determinant is almost exactly equal to an expected value).
- [ ] Print both matrices when the dimension is reasonably small. Currently it's still a little difficult to get a feel for exactly what is wrong in some cases involving small matrices.

Other improvements:
- [ ] Consider letting users provide their own comparators (low-priority unless we get requests for this).
- [ ] Avoid `Copy` trait bounds and avoid storing copied elements instead of references.

PR #163, opened 2017-03-27
Implement assert_scalar_eq!

Andlon:
This PR provides an implementation of a `assert_scalar_eq!` macro which works analogously to `assert_matrix_eq!` and `assert_vector_eq!`.

Currently, we have a bunch of tests for scalars that look roughly like this:

```
let det = lu.det();
let expected_det = 3.0;
let diff = det - expected_det;
assert!(diff.abs() < 1e-6);
```

This is less than ideal for several reasons. For one, if the test fails, the test does not report what caused it to fail. Moreover, the lack of a standardized procedure means that the requirements are often arbitrary. The macro provided in this PR lets us instead do

```
assert_scalar_eq!(lu.det(), 3.0, comp = float);
```

Here we use the float comparator, which provides reasonable defaults, and the tolerance used (ULP and absolute epsilon value) can be tweaked if necessary. Of course, the other comparators (`exact`, `abs` and `ulp`) are also available.

EDIT: This addresses one of the points mentioned in #95.


AtheMathmo:
Looks good to me!

From what I can see this PR does not address areas of the PR where we want to use this new macro. I think we should spin this out as a new issue (tagged easy) and hopefully somebody else picks it up! :)

src/macros/assert_scalar_eq.rs
```diff
@@ -0,0 +1,325 @@
...
+/// assert_scalar_eq!(x, y, comp = abs, tol = 0.1);
+/// # }
+/// ```
+#[macro_export]
+macro_rules! assert_scalar_eq {
+    ($x:expr, $y:expr) => {
+        {
+            // Note: The reason we take slices of both x and y is that if x or y are passed as references,
+            // we don't attempt to call elementwise_matrix_comparison with a &&BaseMatrix type (double reference),
+            // which does not work due to generics.
```
Andlon: Some copypasta comment here which doesn't make sense in the scalar situation

Andlon:
@AtheMathmo: I agree with the point about documentation. I will change this.

Andlon:
OK, I added some very brief notes on where to find details about floating-point comparison issues and how to resolve them with the `assert_matrix/vector/scalar_eq` macros. I guess that should do it?

AtheMathmo:
Look good to me. Merging now and will open a new issue for using these macros.

PR #165, opened 2017-04-05
QR rewrite

Andlon:
This PR builds on #163, so that one should probably be reviewed/merged first, after which I will rebase this one.

The following PR implements the following new functionality:

- A new decomposition struct, `HouseholderQr`, which supersedes `qr_decomp`. Before being unpacked, the orthogonal matrix `Q` is implicitly represented in a compact format.
- The implicit `Q` can be accessed through `HouseholderQr::q()`, which returns an instance of `HouseholderComposition`. This can further be used to compute e.g. `QA`. We will need some more functionality here in the future, most notably left-multiplication with vectors and computing the transposed product `Q^T A`.
- The decomposition can be unpacked to retrieve the full `Q` (m x m) and `R` (m x n) factors for the m x n matrix `A`. One can also unpack the "thin QR" decomposition by calling `unpack_thin()`.
- The way the decomposition is stored internally *should* be compatible with LAPACK, though further work would be required to verify this. I've elected not to expose the internal data at this point.

This PR does *not* implement any special functionality for e.g. least squares or solving linear systems of equations with the QR decomposition. I'd rather leave this for a future PR (and probably discuss in an issue first).

I've spent a little time on performance profiling. It seems the new QR decomposition is very much substantially faster than the existing `qr_decomp()` function, at least on my laptop (compiled with `-C target-cpu = native`):

```
test linalg::qr::householder_qr_decompose_100x100                               ... bench:     250,048 ns/iter (+/- 741)
test linalg::qr::householder_qr_decompose_100x500                               ... bench:   1,474,092 ns/iter (+/- 47,756)
test linalg::qr::householder_qr_decompose_500x100                               ... bench:   2,396,875 ns/iter (+/- 10,293)
test linalg::qr::householder_qr_decompose_unpack_100x100                        ... bench:     508,348 ns/iter (+/- 1,193)
test linalg::qr::householder_qr_decompose_unpack_100x500                        ... bench:   1,741,763 ns/iter (+/- 7,464)
test linalg::qr::householder_qr_decompose_unpack_500x100                        ... bench:  17,703,638 ns/iter (+/- 162,303)
test linalg::qr::householder_qr_decompose_unpack_thin_100x100                   ... bench:     528,069 ns/iter (+/- 20,865)
test linalg::qr::householder_qr_decompose_unpack_thin_100x500                   ... bench:   1,736,694 ns/iter (+/- 4,565)
test linalg::qr::householder_qr_decompose_unpack_thin_500x100                   ... bench:   4,870,457 ns/iter (+/- 33,474)
test linalg::qr::qr_decomp_100x100                                              ... bench:  30,567,430 ns/iter (+/- 119,934)
test linalg::qr::qr_decomp_100x500                                              ... bench:  91,515,084 ns/iter (+/- 10,663,935)
```

In the above benchmarks, `decompose` benches only the `decompose` step, `decompose_unpack` benchmarks `decompose` plus `unpack`, and `decompose_unpack_thin` benchmarks `decompose` plus `decompose_thin`.

`decompose_unpack` would be the equivalent functionality to `qr_decomp`.

It took me some time to land on the current implementation. Particularly because the algorithm presented in Golub & Van Loan breaks down in certain corner cases, so I needed to take some time to figure out the best course of action. I'm pretty happy with the current implementation.

I added some functionality to `Column` in the process. It was useful for this PR, and I figured it could also be useful as part of the public API. Still, I can move these out if you don't want these methods in the API.

I also have to say that for QR decomposition it's a little unfortunate that our matrices are row-major: if `m >> n`, it might be that storing/loading the Householder vectors in/from the columns of the matrix will dominate the runtime - at least for small `n`.


AtheMathmo:
Thank you for this! The benchmarks are awesome and it looks like you got in a lot of other really nice changes at the same time.

I'm terribly busy for the next 1-2 weeks. With some luck I'll be able to review #163 and this PR over the weekend but failing that I promise I'll get around to them after my busy period is done!

Andlon:
There's no rush. Please take as much time as you need! :-) I'll probably be fairly busy the next week (or possibly 2 next weeks) myself.

src/internal_utils.rs
```diff
@@ -19,10 +19,83 @@ pub fn nullify_upper_triangular_part<T, M>(matrix: &mut M)
...
+        axpy(x[j], a_j, y);
+    }
+}
+
+// Given scalar `a` and vectors `x` and `y` of same length, computes the
+// scalar-vector product `y <- y + a * x`.
+//
+/// This function is a stopgap solution until we have a more proper
+/// BLIS/BLAS-like API.
+pub fn axpy<T>(a: T, x: &[T], y: &mut [T])
```
AtheMathmo: Have you compared this to `utils::in_place_vec_bin_op`? It looks like they should be equivalent - if not I'd be interested if you know why!

src/matrix/decomposition/householder.rs
```diff
@@ -0,0 +1,476 @@
...
+/// the implementation compliant with LAPACK.
+pub struct HouseholderReflection<T> {
+    v: Vector<T>,
+    tau: T
+}
+
+impl<T: Float> HouseholderReflection<T> {
+    /// Compute the Householder reflection which will zero out
+    /// all elements in the vector `x` except the first.
+    pub fn compute(x: Vector<T>) -> HouseholderReflection<T> {
```
AtheMathmo: I'm not sure about this name, it's a little too generic, but I can't think of anything else that makes more sense...

src/matrix/decomposition/householder.rs
```diff
@@ -0,0 +1,476 @@
...
+                v: x,
+                tau: T::zero()
+            }
+        }
+    }
+
+    /// Left-multiplies the given matrix by this Householder reflection.
+    ///
+    /// More precisely, let `H` denote this Householder reflection matrix,
+    /// and let `A` be a dimensionally compatible matrix. Then
```
AtheMathmo: I that using `A` when the matrix is `M` in the function signature adds some confusion.

Additionally I'm a little confused by the need to give a buffer. As we have such strict size requirements on the buffer would it not be more appropriate to just allocate this buffer ourselves in the function? I know that this function signature may have performance benefits in some cases but they seem rare enough that I have to question whether it is worth the added complexity to use and understand. Of course - you know this domain a lot better than I so I'll go with your judgement!

src/matrix/decomposition/qr.rs
```diff
@@ -1,17 +1,330 @@
...
+///
+/// where `R1` is an `m x m` upper triangular matrix and
+/// `R2` is an `m x (n - m)` general matrix.
+///
+/// To actually compute the QR decomposition, see
+/// [Householder QR](struct.HouseholderQr.html).
+#[derive(Debug, Clone)]
+pub struct QR<T> {
+    /// The orthogonal matrix `Q` in the decomposition `A = QR`.
+    pub q: Matrix<T>,
```
AtheMathmo: I'm not sure why these fields are public. This would allow the user to construct invalid `QR` and `ThinQR` structs, i.e.:

```rust
let qr = QR { q: Matrix::zeros(1,1), r: Matrix::ones(2,2) }
```

From what I can see this doesn't cause any issues at the moment. These structs are only for storage if a user wants to unpack. However, if in future we want to work with the unpacked structs directly this may be a problem.

It's not a big deal but I think restricting the construction here is a little safer. We should then move the field descriptions into the struct descriptions.

src/matrix/decomposition/qr.rs
```diff
@@ -1,17 +1,330 @@
...
+    let q_operator = householder::create_composition(qr, tau);
+    q_operator.first_k_columns(m)
+}
+
+fn extract_r1<T: Float>(qr: &Matrix<T>) -> Matrix<T> {
+    let m = qr.rows();
+    let n = qr.cols();
+    let mut data = Vec::<T>::with_capacity(m * n);
+
+    assert!(m > n, "We only want to extract r1 if m > n!");
```
AtheMathmo: VERY minor: As this function is only used internally perhaps this should be a `debug_assert`?

src/matrix/mod.rs
```diff
@@ -396,3 +396,85 @@ fn forward_substitution<T, M>(l: &M, y: Vector<T>) -> Result<Vector<T>, Error>
...
 }
+
+impl<'a, T> ColumnMut<'a, T> where T: Clone {
+    /// Clones the elements of the given slice of compatible size
+    /// into this column.
+    ///
+    /// # Panics
+    /// - The slice does not have the same length as
+    ///   the number of rows in the column.
+    pub fn clone_from_slice(&mut self, slice: &[T]) {
```
AtheMathmo: For consistency we may want to add similar functions for `Row`/`RowMut`? I don't think these exist yet.

Andlon: Agree! Perhaps defer it to a separate issue/PR after eventually merging this one though?

src/matrix/decomposition/qr.rs
Andlon: I sort of agree, but I think it's better for the user to have his program panic if we've somehow implemented this in the wrong way than the user having to deal with potentially subtly wrong behavior. Then it's an internal error which the user can submit to us, and we can quickly pinpoint the problem. What do you think?

src/matrix/decomposition/qr.rs
Andlon: The reason is so that we can have the syntax:

```rust
let QR { q, r } = qr.unpack();
```

In other words, the `QR` and `ThinQR` structs are only used for output from `unpack` and `unpack_thin`. We should never take these as input anywhere (then it's better to take the corresponding `HouseholderQr` struct) for the reasons you mention.

The thing is that we want to hand off the `q` and `r` matrices to the user, which leaves us with either something like this `QR` approach, or we have to return a tuple `(q, r)`. I'm not in favor of the latter because of ordering. For QR, this is not such a big problem, because the ordering is very obvious, but for e.g. LU decomposition one may permute the ordering of the parameters, e.g. `(p, l, u)` or `(l, u, p)`. Thus this is consistent with the approach used by `PartialPivLu` and `FullPivLu`, which both return `LUP` and `LUPQ` structs.

src/matrix/decomposition/householder.rs
Andlon: It's fairly generic indeed, but consider the usage: `HouseholderReflection::compute(vec)`. Basically, compute the householder reflection associated with the vector `vec`. So it's not too bad I guess. But definitely open for suggestions for a better name - I couldn't think of anything better myself! In any case, please note that `HouseholderReflection` is not exposed in the public API at this point - it's only used internally (though `HouseholderComposition` is part of the public API).

src/matrix/decomposition/householder.rs
Andlon: Note that `M` is the *type* of the matrix, so I don't think this is problematic - but I agree that the parameter, which is currently `matrix`, could perhaps be named similarly to the documentation string. Due to snake_case, that would mean `a`, I guess.

By the way, I've been contemplating using `#[allow(non_snake_case)]` selectively in parts of the rulinalg API - this would let us actually use big letters for matrices, which reads much nicer from the mathematical standpoint. I haven't decided whether this is a good idea or not yet.

src/matrix/decomposition/qr.rs
AtheMathmo: Thanks for reminding me :)

src/internal_utils.rs
Andlon: Huh, I think I looked at it before implementing these BLAS-like routines. Early on it wasn't quite the same, but I guess in the end I ended up with the exact same. Interesting! `axpy` could certainly have been implemented in terms of `in_place_vec_bin_op`.

I don't mind changing it, but on the other hand it's actually a little easier to reason about exactly what's going on without using `utils::in_place_vec_bin_op`, because `axpy` is such a common mathematical operation and it's convenient for its implementation to mirror the mathematical definition (which gets slightly hidden if we use `in_place_vec_bin_op`). I don't know, I don't mind either way, so let me know your preference :-)

src/matrix/decomposition/householder.rs
Andlon: Sorry, forgot to respond to the buffer part. The reason is that this particular function is called over and over and over again during the QR decomposition and unpacking.

Specifically, given an `m x n` matrix `A`, the `HouseholderQr::decompose` function calls this function `n` times. When unpacking, the function is called another `n` times.

This represents a lot of unnecessary memory allocations if a new buffer is allocated every time it's called. On a modern computer, this doesn't make a noticeable difference (I benched it), because allocation is fast and cheap compared to the act of actually computing the rank-1 update and the transposed matrix-vector product, but for, say, a resource-constrained environment where allocations are slow this might matter. I know that we shouldn't worry too much about these scenarios as they're perhaps a bit niche, but I just have this stance that any linear algebra library worth its salt should try to minimize spurious and unnecessary memory allocations if it's reasonably possible to do so. Because linear algebra is such a basic facet of modern computing, I feel that we shouldn't make too many assumptions about the hardware the user is running on.

Note that this function is also not part of the public API of `rulinalg`, which I think makes the pattern here somewhat more acceptable. However, I can also rewrite it to allocate its own buffer if you prefer this. I also want to say that I don't think this would have been an acceptable function signature if this were part of the public API!

src/internal_utils.rs
AtheMathmo: I see! I think it is ok to leave it as-is. It _might_ be worth using `axpy` as an alias to a `in_place_vec_bin_op`, just to cut down on code. But this isn't really important. You should stick to whatever is most comfortable for you - these are only placeholders and you will be doing the most work on this code in the near future.

src/matrix/decomposition/householder.rs
AtheMathmo: I agree with you - it makes sense and I can't think of anything better :).

src/matrix/decomposition/householder.rs
AtheMathmo: Sorry you're right about the type, silly oversight on my part. I think the docs are actually ok as they are - I was just a little tired when reading them the first time around. Personally I'd be against breaking snake casing. It helps us get a little closer to mathematical notation but we'll never be able to match it. I think our current balance of rusty design and linear algebra conventions are pretty good. Of course, we may want to revisit this in future.

Thanks for clearing up the buffer issue. I agree that this approach is definitely the one we want. I had overlooked a few things when I wrote my previous comment :). Even in the public API I think it is good to give users the option to write more complex, high performance code. We should definitely keep this as it is currently.

src/matrix/decomposition/qr.rs
AtheMathmo: I think you're right. This isn't an expensive check and I'd rather our users help us to discover our bugs, even in release environments :)

src/matrix/mod.rs
AtheMathmo: Sure!

We could even have these become functions in the `BaseMatrix`/`BaseMatrixMut` traits. These would also make sense for matrices I think and maybe capture some rarely needed functionality.

Andlon: Well, that depends - do we want to introduce column-major matrices soon? I think we need to (and we can without much problems once we have the BLAS-like API in place), in which case it would be ambiguous what this method would do for an arbitrary slice

src/matrix/decomposition/householder.rs
Andlon: Agree with your assessment about snake casing - it would just be a minor step forward at the cost of breaking a standard Rust convention, which doesn't seem worth it!

src/internal_utils.rs
Andlon: Sure, I can make that change!

AtheMathmo: Note though that this depends on whether there is any performance advantage to the explicit `axpy` routine you've written :)

Andlon:
Thanks for the review! You asked some important questions about some of the design choices I've made, and we've seem to have resolved most of it already. There are a couple of minors things I'll still have a look at. I should be able to do so during the next few days, possibly even tomorrow.

Note that since #163 was merged, I've rebased this branch on top of `master` again (since it already included the functionality from the #163 branch).

I'm not quite sure what to do about the inconsistent naming of the structs. This is actually also the case for `LUP` vs `PartialPivLu`. I thought `HouseholderQr` looked a bit silly, but it was consistent with Rust naming conventions (I think?), so that's why I chose it. For the `QR` struct, I thought it was better to have the name as accurately as possible reflect the decomposition, in which case `QR `would be better than `Qr`.

In any case, I don't have a strong opinion either way. If you have a preference, please let me know :-)

AtheMathmo:
I also don't have a strong opinion. I think I prefer to capitalize the acronyms, i.e. `PartialPivLU` and `HouseholderQR`. However, given that we have already set the trend with previous work I say we keep this as written currently for now.

We can think about changing it in future if it is annoying for enough people :)

Andlon:
I rewrote `axpy` in terms of `in_place_vec_bin_op`, and there is no performance difference. In fact, it performed slightly better - which I think must come down to some background processes running on my computer, because the two pieces of code are equivalent to me (at the very least the previous `axpy` implementation shouldn't have been slower!).

I've gone through your comments again and I can't find anything else that needs to be changed. Let me know if I've missed something!

src/matrix/mod.rs
AtheMathmo: I think that this is the only thing left to address.

I'm not ready to make a decision on column-major matrices just yet. It seems to me to be a little way off (after completing BLAS-like work at least). I'm happy to leave these functions only for `Row` and `Col` types for now (with `Row` coming later).

AtheMathmo:
I'll take another quick look through later today so we can get this merged - but I think it is good!

PR #167, opened 2017-04-11
Fix and test metric of two MatrixSlice's (#154)

### LINKED ISSUE #154: Cannot compute induced metric of two `MatrixSlice`s
The current trait bounds restrict us taking the metric of two slices. It looks like this:

```rust
impl<'a, 'b, U, T, M1, M2> MatrixMetric<'a, 'b, T, M1, M2> for U
    where U: MatrixNorm<T, M1>,
    M1: 'a + BaseMatrix<T>,
    M2: 'b + BaseMatrix<T>,
    &'a M1: Sub<&'b M2, Output=M1> {

    fn metric(&self, m1: &'a M1, m2: &'b M2) -> T {
        self.norm(&(m1 - m2))
    }
}
```

The issue is the last line of the `where` clause. We require that the output be of type `M1` but this is not the case for slices. I think we can remedy this by requiring that the Output is a `Matrix<T>`.

initiumSrc:
This resolves #154 with the suggested solution. I also added MatrixSlice specific tests.

Is `::matrix::Matrix<T>` acceptable or is a `use matrix::Matrix` statement desired? Felt unnecessary as it's only used 'once'.

Also, is it OK to split the matrixslice tests into their own test functions or do I want to move the slice tests into the matrix tests?

### DIFF
--- a/src/norm/mod.rs
+++ b/src/norm/mod.rs
@@ -235,6 +235,12 @@ mod tests {
         let m = matrix![3.0, 4.0;
                         1.0, 3.0];
         assert!((MatrixNorm::norm(&Euclidean, &m) - 35.0.sqrt()) < 1e-14);
+    }
+
+    #[test]
+    fn test_euclidean_matrix_slice_norm() {
+        let m = matrix![3.0, 4.0;
+                        1.0, 3.0];
 
         let slice = MatrixSlice::from_matrix(&m, [0,0], 1, 2);
         assert!((MatrixNorm::norm(&Euclidean, &slice) - 5.0) < 1e-14);
--- a/src/norm/mod.rs
+++ b/src/norm/mod.rs
@@ -285,7 +285,7 @@ mod tests {
     }
 
     #[test]
-    fn test_euclidean_matrixslice_metric() {
+    fn test_euclidean_matrix_slice_metric() {
         let m = matrix![
             1.0, 1.0, 1.0;
             1.0, 1.0, 1.0;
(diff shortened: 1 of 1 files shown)


src/norm/mod.rs
```diff
@@ -285,6 +285,48 @@ mod tests {
     }
 
     #[test]
+    fn test_euclidean_matrixslice_metric() {
```
Andlon: Super nit-picky: Since `MatrixSlice` usually becomes `matrix_slice` in snake case per usual camel case -> snake case conversion conventions. I think that it doesn't really need to be changed unless there are also other changes that are necessary to make though!

initiumSrc: I'll have that changed 👍 

Also, should the matrix slice norm test reside inside the test_euclidean_matrix_norm? I reckon it should be extracted to test_euclidean_matrix_slice_norm if changes were to hit either of the implementations the current test would not explicitly pinpoint the source.

Andlon: I agree with your assessment that it's better to have these tests separate as they essentially test two different implementations.

In general, tests in rulinalg are often not as granular as they should be - I'd say this is largely due to the fact that the default unit testing facilities of rust/cargo are very rudimentary. So things like sharing data between independent tests is often difficult so we sometimes "cheat" by putting multiple blocks within a single `#[test]`.

AtheMathmo:
I think this PR does what it should, thanks @initiumSrc !

As @Andlon pointed out the design of the metric traits are a little unfortunate. I guess we should revise and improve later but this PR fixes some immediate issues.

PR #168, opened 2017-04-14
Add Slice* for Vector

### LINKED ISSUE #19: Adding a slice equivalent for `Vector`.
It would be good to add a slicing equivalent for Vector (similar to `MatrixSlice` for `Matrix`).

For `Vector ` we can probably do this in a nicer way than `Matrix`. We can implement a fat pointer type to represent the underlying data in a `Vector` - and then implement the relevant methods on a reference to this type (similar to the `Vec` - `&[T]` relationship in std).

Using `Deref`/`Borrow`/`AsRef` we can cut down on code by implementing all functions taking `&self` on the `VectorSlice` type.

It would be acceptable to simply mirror the current behaviour for matrices - but we should try to get the above working.

c410-f3r:
Fixes #19. Tried to mimic most tests from the original implementation but figured out that all news tests must be reformulated to a new format, something like the tests for `MatrisSlice*`. Afterwards I will create another PR to handle this problem.
`BaseVector*` and `VectorSlice*` are different and simpler than `BaseMatrix*` and `MatrixSlice*`, better look this area with extra care.
I split the operators implementations in several mods to preserve my mental sanity, don't know if you guys want to keep things separated in this way.


Andlon:
@c410-f3r: thanks for putting into the effort for this issue! I'm somewhat busy at the moment so I only had time for a brief look. Looks good from a cursory glance, but I will try to make time for a detailed review, hopefully next week.

Andlon:
I'll try to start looking into this tomorrow. It's a pretty big PR and I'm not overly familiar with the existing code from which this has been based on, so it might take me some time. Sorry for the delay so far!

c410-f3r:
Relax, you don't need to rush things. Thanks for the attention!

src/vector/base.rs
```diff
@@ -0,0 +1,547 @@
...
+    /// let a = vector![1, 2, 3, 4, 5];
+    /// let b = a.as_slice();
+    /// # }
+    /// ```
+    fn as_slice(&self) -> VectorSlice<T> {
+        unsafe { VectorSlice::from_raw_parts(self.as_ptr(), self.size()) }
+    }
+
+    /// Horizontally concatenates two vectors. With self on the left.
+    ///
```
Andlon: My understanding of vectors in rulinalg is that they represent column vectors - in which case horizontally concatenating might be a little misleading...?

src/vector/base.rs
```diff
@@ -0,0 +1,547 @@
...
+    {
+        let mut new_data = Vec::with_capacity(self.size() + vector.size());
+
+        new_data.extend_from_slice(self.data());
+        new_data.extend_from_slice(vector.data());
+
+        Vector::new(new_data)
+    }
+
+    /// Returns the slice that represents the currently struct view.
```
Andlon: Perhaps rephrase this a bit. Something like "Returns a slice representing the underlying data of the vector"?

src/vector/base.rs
```diff
@@ -0,0 +1,547 @@
...
+    /// assert_eq!(*a.get(0).unwrap(), 1)
+    /// # }
+    /// ```
+    fn get(&self, index: usize) -> Option<&T> {
+        if index >= self.size() {
+            None
+        } else {
+            unsafe { Some(self.get_unchecked(index)) }
+        }
+    }
```
Andlon: This can also be implemented in terms of `self.data().get(index)` (and thus we avoid a block of `unsafe`).

src/vector/base.rs
```diff
@@ -0,0 +1,547 @@
...
+    /// let b = &a.select(&[0, 3]);
+    ///
+    /// assert_eq!(b.size(), 2);
+    /// assert_eq!(b.data(), &[1, 4]);
+    /// # }
+    /// ```
+    ///
+    /// # Panics
+    ///
+    /// - Panics if indixes exceed the vector size.
```
Andlon: Typo: indexes -> indices (also in the variable name)

src/vector/base.rs
```diff
@@ -0,0 +1,547 @@
...
+    /// assert_eq!(c, 5.0);
+    /// # }
+    /// ```
+    fn norm<N>(&self, norm: N) -> T
+        where N: VectorNorm<T, Self>
+    {
+        norm.norm(self)
+    }
+
+    /// Select block vector from vector.
```
Andlon: Perhaps rephrase this

src/vector/base.rs
```diff
@@ -0,0 +1,547 @@
...
+            }
+        }
+
+        Vector::new(values_vector)
+    }
+
+    /// Size of the vector.
+    fn size(&self) -> usize;
+
+    /// Split the vector at the specified index returning two `VectorxSlice`s.
```
Andlon: typo: VectorxSlice

src/vector/base.rs
```diff
@@ -0,0 +1,547 @@
...
+    /// Get a reference to an element in the vector without bounds checking.
+    unsafe fn get_unchecked(&self, index: usize) -> &T {
+        &*(self.as_ptr().offset(index as isize))
+    }
+}
+
+/// Trait for mutable vector structs.
+pub trait BaseVectorMut<T>: BaseVector<T> {
+    /// Left index of the vector.
+    fn as_mut_ptr(&mut self) -> *mut T;
```
Andlon: "Left" is a bit of a strange description for a column vector ;-)

src/vector/base.rs
```diff
@@ -0,0 +1,547 @@
...
+    ///
+    /// let mut a = vector![1, 2, 3, 4, 5];
+    /// let b = a.as_mut_slice();
+    /// # }
+    /// ```
+    fn as_mut_slice(&mut self) -> VectorSliceMut<T> {
+        unsafe { VectorSliceMut::from_raw_parts(self.as_mut_ptr(), self.size()) }
+    }
+
+    /// Returns the mutable slice that represents the currently struct view.
```
Andlon: Same as before, perhaps rephrase this a little

src/vector/base.rs
```diff
@@ -0,0 +1,547 @@
...
+    /// assert_eq!(*a.get_mut(0).unwrap(), 2);
+    /// # }
+    /// ```
+    fn get_mut(&mut self, index: usize) -> Option<&mut T> {
+        if index >= self.size() {
+            None
+        } else {
+            unsafe { Some(self.get_unchecked_mut(index)) }
+        }
+    }
```
Andlon: Think we can also reimplement this in terms of `self.data()`

Andlon:
Oh, and, I have to ask you for some patience - it might take some time to get this merged. There's a number of unmerged PRs up for merging at the moment, and the order in which this gets done is not so clear to me at the moment. Let's see what @AtheMathmo thinks later on :-)

src/norm/mod.rs
```diff
@@ -28,48 +28,44 @@
...
+pub trait MatrixNorm<M, T>
+    where M: BaseMatrix<T>
+{
     /// Computes the matrix norm.
     fn norm(&self, m: &M) -> T;
 }
 
 /// Trait for matrix metrics.
-pub trait MatrixMetric<'a, 'b, T, M1: 'a + BaseMatrix<T>, M2: 'b + BaseMatrix<T>> {
+pub trait MatrixMetric<'a, 'b, M1, M2, T>
```
c410-f3r: Hehe =P
That was a whim of mine because I like things arranged in alphabet/crescent order.
Ok, I will undo this arrangement.

src/vector/base.rs
c410-f3r: Indeed

src/vector/base.rs
c410-f3r: Coded lots of lines with `indexes`, is there a convention to use `indexes` or `indices`?
As my mother tongue is derived from Latin, I think it's much easier to use `indices` instead of its Anglo-Saxon counterpart, but `indexes` sounds better when coupled with the `index` singular form.

src/vector/base.rs
c410-f3r: Are you sure? I made the whole thing thinking about row vectors :(

c410-f3r:
TODO:

- [ ] Undo matrix norm arrangement.
- [ ] Check generated documentation.

src/vector/base.rs
Andlon: Oh, I actually didn't know "indexes" was a valid spelling - I've never seen it written. At least particularly in software- or mathematical literature `indices` seems to be by far the most common form. That's my impression anyway!

src/vector/base.rs
Andlon: Quite sure! Consider the fact that we routinely compute matrix-vector products `Ax` - if `x` were a row vector this wouldn't in general make sense!

From what I saw in your code, it doesn't make any difference - only a few spots in the documentation I think?

src/vector/base.rs
c410-f3r: It then shall be `indices`

src/vector/base.rs
```diff
@@ -0,0 +1,547 @@
+//! Traits for vector operations.
+//!
+//! These traits defines operations for structs representing vectors.
+//!
+//! Implementations are provided for
+//! - `Vector`: an owned vector
+//! - `VectorSlice`: a borrowed immutable block of `Vector`
+//! - `VectorSliceMut`: a borrowed mutable block of `Vector`
```
c410-f3r: Couldn't test with generated doc because module is private but tested with online markdown editor and things where OK.

src/vector/base.rs
c410-f3r: Guess you are right, it doesn't make any difference!

PR #169, opened 2017-04-14
Formatting

c410-f3r:


Andlon:
I appreciate that you're making a push for automated formatting, @c410-f3r! I'm excited to make this part of our workflow. As with #168, I will try to take a look at this soon, hopefully next week!

.travis.yml
```diff
@@ -1,13 +1,18 @@
...
-  - cargo build --features io
-  - cargo test --features io
\ No newline at end of file
+- |
+  export PATH=$PATH:~/.cargo/bin &&
+  cargo fmt -- --write-mode=diff &&
+  cargo build --verbose &&
+  cargo test --verbose &&
+  cargo build --features io &&
+  cargo test --features io
```
Andlon: Should probably be a newline at the end here?

Script otherwise looks sensible to me!

src/matrix/decomposition/lu.rs
```diff
@@ -816,10 +816,19 @@ mod tests {
...
+
+            1.66666666666666630e-01,  6.66666666666666519e-01,
+            -6.66666666666666519e-01, 4.99999999999999833e-01;
+
+            -3.88888888888888840e-01, 1.11111111111111174e-01,
+            5.55555555555555358e-01,  -4.99999999999999833e-01;
+
+            7.40740740740740838e-02, -9.25925925925925819e-01,
+            3.70370370370370294e-01, 5.13992141030165006e-17
+        ];
```
c410-f3r: `#[cfg_attr(rustfmt, rustfmt_skip)]` won't help in this case, take a look in this [issue](https://github.com/rust-lang-nursery/rustfmt/issues/1298).
We could increase line width to 120 with the `max_width` parameter.

src/matrix/base/mod.rs
```diff
@@ -1684,8 +1691,8 @@ pub trait BaseMatrixMut<T>: BaseMatrix<T> {
                                                              self.rows(),
                                                              mid,
                                                              self.row_stride());
-                    slice_2 = MatrixSliceMut::from_raw_parts(self.as_mut_ptr()
-                                                                 .offset(mid as isize),
+                    slice_2 = MatrixSliceMut::from_raw_parts(self.as_mut_ptr().offset(mid as
+                                                                                      isize),
```
c410-f3r: Maybe related to [#1405](https://github.com/rust-lang-nursery/rustfmt/issues/1405). I guess we can't let rustfmt handle all cases automatically.

src/matrix/decomposition/lu.rs
```diff
@@ -423,17 +425,18 @@ impl<T: 'static + Float> FullPivLu<T> {
         }
 
         Ok(FullPivLu {
-            lu: lu,
-            p: p.inverse(),
-            q: q.inverse()
-        })
+               lu: lu,
+               p: p.inverse(),
+               q: q.inverse(),
```
c410-f3r: Couldn't find a parameter that modifies this behaviour, maybe it is related to [#1019](https://github.com/rust-lang-nursery/rustfmt/issues/1019).

Andlon:
@c410-f3r: if lines are too long, will this fail CI? That is, will tests on Travis CI fail if there are lines which are too long, or will `rustfmt` just spit out a warning?

I think in some instances it's preferable to have lines that are too long than to actually break them up. We might be able to live with this if `rustfmt` doesn't cause our CI to fail in these cases.

c410-f3r:
Didn't make any test but according to [this](http://kneit.in/2016/11/26/rustfmt-in-travisci.html), it will fail. Since `#[cfg_attr(rustfmt, rustfmt_skip)]` won't work in this case, the only way is increase line width to 120.

Maybe wait for 1.0 release of rustfmt and see if things are more stable?

Issue #147, opened 2017-02-10
Remove `Any` type bound throughout library


AtheMathmo:
We had to place an `Any` type bound because of [TypeId](https://doc.rust-lang.org/std/any/struct.TypeId.html) needed in the matrix multiplication module. Now the `Any` bound has been replaced with a static lifetime bound and we should do the same.

PR #171, opened 2017-04-28
Issue #147

c410-f3r:
Solves #147.


AtheMathmo:
Thanks! At a glance this looks good. Will have to check it out a little more locally, but should be able to merge quite quickly.

Andlon:
I labeled this as a breaking change, but it's a very mild breaking change. We might still want to push it back to next "major version", i.e. 0.5. We might also want to release 0.5 very soon, and push back the name change to 0.6.

Issue #143, opened 2017-01-31
Categories for rulinalg


Andlon:
I just noticed through [this reddit link](https://www.reddit.com/r/rust/comments/5r72aj/cratesio_has_categories/?ref=share&ref_source=link) that [crates.io](https://crates.io) has support for [categories](https://crates.io/categories), although this does not seem to be accessible from the front page yet. I suspect this is just a matter of time.

In any case, I looked through the available categories, and I think Rulinalg should be present in the "Science" category. I think this is just a matter of adding 

```
categories = [ "science" ]
```

to the `[package]` section of `Cargo.toml`.

I would make a PR myself, but I'm not on my dev computer, and I really should be doing other things right now anyway, heh. Just wanted to note it down before I forgot. Maybe I'll get it done later.

PR #173, opened 2017-04-28
Add sciency category

c410-f3r:
Solves #143.
Couldn't find another category that fits `rulinalg`, maybe `algorithms`?


AtheMathmo:
I think just science is good for now. Thanks!

Will merge after I confirm this does match the cargo spec needed (it looks fine at a glance).

Andlon:
I checked the cargo spec, and it matches. Ready to merge, as far as I can see!

Issue #175, opened 2017-04-28
Using assert_*_equal macros in all tests


AtheMathmo:
We have just merged #163 which gives us flexible scalar comparisons. Currently many of our tests have checks like this:

```rust
let det = lu.det();
let expected_det = 3.0;
let diff = det - expected_det;
assert!(diff.abs() < 1e-6);
```

Instead we can use the new macro:

```rust
assert_scalar_eq!(lu.det(), 3.0, comp = float);
```

This is cleaner and more flexible. To resolve this issue we should convert all existing comparison checks to use the `assert_*_eq!` macros.

Andlon:
I believe this was resolved by #177, so closing this issue.

PR #177, opened 2017-05-05
Fixed (#175) and cleaned up a few tests

mark-blum:
Changed all the easy assert!(diff < eps) to assert_scalar_eq! as I could find

There are still a few instances of a regular diff < eps that can be
found in the src/matrix/decomposition/eigen.rs, but they are inside
iterators that make it harder to easily change.

I also spruced up some of the tests, to remove unneeded clones or things of the sort and removed compiler warnings about unused 'uses'

### DIFF
--- a/src/matrix/decomposition/eigen.rs
+++ b/src/matrix/decomposition/eigen.rs
@@ -408,9 +408,8 @@ mod tests {
         let v1 = vector![eigenvecs[[0, 0]], eigenvecs[[1, 0]]];
         let v2 = vector![eigenvecs[[0, 1]], eigenvecs[[1, 1]]];
 
-        let epsilon = 0.00001;
-        assert!((&a * &v1 - &v1 * lambda_1).iter().all(|&c| c < epsilon));
-        assert!((&a * &v2 - &v2 * lambda_2).iter().all(|&c| c < epsilon));
+        assert_vector_eq!(&a * &v1, &v1 * lambda_1, comp = float);
+        assert_vector_eq!(&a * &v2, &v2 * lambda_2, comp = float);
     }
 
     #[test]
--- a/src/matrix/decomposition/eigen.rs
+++ b/src/matrix/decomposition/eigen.rs
@@ -409,8 +409,8 @@ mod tests {
         let v2 = vector![eigenvecs[[0, 1]], eigenvecs[[1, 1]]];
 
         let epsilon = 0.00001;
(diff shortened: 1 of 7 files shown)


tests/mat/mod.rs
```diff
@@ -197,24 +197,17 @@ fn qr() {
...
+    for (q_test, q_true) in q.iter().zip(true_q.iter()) {
+        assert_scalar_eq!(q_test, q_true, comp = float, eps = 1e-6);
     }
-
-    for (val, expected) in r_diff.into_iter().zip(true_r.data().iter()) {
-        assert!(val < tol, format!("diff is {0}, expecting {1}", val, expected));
+    for (r_test, r_true) in r.iter().zip(true_r.iter()) {
+        assert_scalar_eq!(r_test, r_true, comp = float, eps = 1e-6);
     }
-
```
Andlon: These tests are fairly old, and in fact they predate the `assert_*_eq!` macros. I think in this case we can make it even simpler:

```rust
assert_matrix_eq!(q, true_q, comp = abs, tol = 1e-6);
```

which eliminates the need for the loops and makes the error output much more useful! Note that it's possible to use `comp = float` with `eps = 1e-6` as you've done too, but the absolute tolerance is more explicit in the case that you just want an absolute comparison.

mark-blum:
I have simplified that test even further

src/matrix/decomposition/eigen.rs
```diff
@@ -409,8 +409,8 @@ mod tests {
         let v2 = vector![eigenvecs[[0, 1]], eigenvecs[[1, 1]]];
 
         let epsilon = 0.00001;
-        assert!((&a * &v1 - &v1 * lambda_1).into_vec().iter().all(|&c| c < epsilon));
-        assert!((&a * &v2 - &v2 * lambda_2).into_vec().iter().all(|&c| c < epsilon));
+        assert!((&a * &v1 - &v1 * lambda_1).iter().all(|&c| c < epsilon));
```
AtheMathmo: Can we not use `assert_vector_eq!` here?

Andlon: Yeah, definitely! I missed that one, good catch :-)

Andlon:
Hmm, Travis doesn't see this PR for some reason. I don't know why. I guess we have to wait to see if it catches up!

mark-blum:
Yup, we could use `assert_vector_eq!` there. I just push a new commit that changes that

Issue #172, opened 2017-04-28
Make `Matrix::from_fn` row-major


AtheMathmo:
The current function produces a matrix `A_ij = f(j,i)`. This is inconsistent with the rest of the library.

We should instead have the function construct `A_ij = f(i,j)`.

PR #178, opened 2017-05-05
Made Matrix::from_fn row major, fixes (#172)

mark-blum:
Added a test to make sure it is row major as well

I wonder if a better example could be made. The current one is completely symmetrical and doesn't change if row and col get swapped.


Andlon:
Thanks for this PR! From a brief glance it looks good.

I agree about the example. Do you have anything more suitable in mind?

Andlon:
(Just mentioning #172 here so that we get a reverse link)

PR #180, opened 2017-05-07
Use macros in the README examples

### LINKED ISSUE #130: Update README with more enticing example
The README would perhaps look a little more interesting if one uses the `matrix!` macro, as well as the `assert_matrix_eq!` macro in the example at the end of the document.

This is a simple issue that should be suitable for users looking to make their first contribution, or existing contributors looking to tackle a simple issue.

mbattifarano:
Use the `matrix!` and `assert_matrix_eq!` macros in the README examples.

Resolves #130

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,7 @@ rulinalg="0.4.2"
 And then import the library using:
 
 ```rust
+#[macro_use]
 extern crate rulinalg;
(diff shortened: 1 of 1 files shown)


AtheMathmo:
Great, thanks!

PR #183, opened 2017-05-23
Remove Any bounds

theotherphil:
Fixes https://github.com/AtheMathmo/rulinalg/issues/147

### DIFF
--- a/benches/lib.rs
+++ b/benches/lib.rs
@@ -7,14 +7,15 @@ extern crate test;
 extern crate rand;
 
 pub mod linalg {
-	mod iter;
-	mod matrix;
-	mod svd;
-	mod lu;
 	mod cholesky;
-	mod qr;
+	mod iter;
+	mod lu;
+	mod matrix;
 	mod norm;
-	mod triangular;
 	mod permutation;
+	mod qr;
+	mod svd;
+	mod transpose;
+	mod triangular;
 	pub mod util;
 }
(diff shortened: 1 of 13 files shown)


theotherphil:
Is there anything blocking merging this?

Andlon:
Hey @theotherphil, sorry, @AtheMathmo and I have both been very busy as of late.

A few brief comments:

- There is already an existing (unmerged) PR #171 which seems very similar. I haven't looked in detail at the changes, but I expect them to be almost identical.
- This is a breaking change, although a very minor one, so we need to decide if we can put it in a patch version or if we should postpone to the next minor version (though perhaps our next release is a new minor version)
- This PR also includes some functionality for in-place transpose, which is a little confusing since it's a separate issue. I was also lead to believe that this kind of functionality depends on some unavailable functionality in Rust atm, such as 128-bit integers?

theotherphil:
Oops, I didn't mean to include the in-place transpose changes! I'd forgotten about this commit by the time I made the in-place-transform changes.

I hadn't seen the other unmerged commit. I'll close this pull request.

Andlon:
@theotherphil: No worries! Thanks for taking the time to make the PR. I'm sorry that progress on merging contributions for this library has slowed down lately, especially considering the number of interested contributors. I hope to have some more time again myself soon (recently been focused on starting my new job in a new country, including all the paperwork and getting the apartment in order), which should help a little with the situation.

PR #184, opened 2017-05-25
Improve variance computation

### LINKED ISSUE #49: Variance performance improvement
We can get a _slight_ performance gain in the `Matrix::variance` function.

``` rust
let v = Vector::new(t);
variance = variance + &(&v - &mean).elemul(&(&v - &mean));
```

Here we compute `v - mean`  twice. Instead we should compute it once and then modify.

``` rust
// Computes diff using the space allocated for `t`
let diff = Vector::new(t) - &mean;
variance = variance + diff.elemul(&diff);
```

This may well be optimized away already (?) - but we should make it explicit.

mbattifarano:
Closes #49

### DIFF
--- a/src/matrix/impl_mat.rs
+++ b/src/matrix/impl_mat.rs
@@ -304,9 +304,9 @@ impl<T: Float + FromPrimitive> Matrix<T> {
                 }
             }
 
-            let v = Vector::new(t);
+            let diff = Vector::new(t) - &mean;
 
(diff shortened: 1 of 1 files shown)


c410-f3r:
Omg! So many PRs!

Thanks for your contribution, @mbattifarano. I guess @AtheMathmo and/or @Andlon are a bit busy these days but soon they will check out your code.

AtheMathmo:
Looks good, thank you!

PR #189, opened 2017-06-02
Cholesky: Fix misleading error message

vks:
The diagonals of the *decomposed* matrix were negative, which means the original matrix was not positive definite (even the the diagonal was positive).


AtheMathmo:
I think this is a good change from the user perspective! @Andlon am I missing anything here?

Andlon:
Thanks for the PR @vks! Yes, I agree that the original error message is indeed misleading.

Note that the documentation for the method also needs to be updated as it has a similar wording as the original error message. Would you mind changing this as well @vks? In this case, we need to be a bit careful with how it is worded.

I see now that also the other documented error message is misleading, since it states that:

> A diagonal entry is effectively zero to working precision.

As you pointed out, it is not the diagonal entry in the original matrix, but the diagonal entry in the decomposition which is being carried out. Perhaps just replace this by "The matrix is effectively singular to working precision"?

PR #195, opened 2018-03-02
De/serialization with serde

garro95:
This pull request add serialization and deserialization with serde for matrix and vector and related tests


garro95:
Serde is enabled with the "serde" feature and also related tests are run with the feature. travis.yml should be updated accordingly

